Issue #45, opened 2022-08-22
Ability to export specific resources from a Workspace using the preset-cli


Vitor-Avila:
### Motivation
Currently, it's possible to use the `preset-cli` to export all assets from a Workspace. However, in some cases users might just want to export modified assets, or specific ones.

The CLI only supports exporting all assets from the source Workspace.

### Proposed solution
Implement the ability to specify which assets should be exported. A possible solution would be to specify the `IDs` to be exported.

betodealmeida:
Closed this by mistake.

betodealmeida:
https://github.com/preset-io/backend-sdk/pull/84 allows specifying asset type

PR #85, opened 2022-09-27
feat: export resources by ID

betodealmeida:
Allow specifying resource ID when exporting. Eg:

```bash
$ superset-cli http://localhost:8088/ export --asset-type=dashboard --dashboard-ids=1,2,3
```

Will only export dashboards with IDs 1, 2, and 3.


betodealmeida:
Closes https://github.com/preset-io/backend-sdk/issues/45.

PR #88, opened 2022-09-28
chore: improve export by ID

betodealmeida:
This PR introduces a few changes to the `superset export` command. First, if IDs are specified for any resource type, then all assets are exported by ID. The command:

```bash
superset-cli export --database-ids=1,2
```

Will export only databases with ID 1 and 2, and no datasets, charts, nor dashboards will be exported. **Before:** this would export databases with ID 1 and 2, but also **all** datasets, charts, and dashboards.

Second, when an ID is specified, related assets are also exported. The command:

```bash
superset-cli export --dashboard-ids=1 --asset-type=dashboard
```

Will export dashboard with ID 1, but also all its associated charts, datasets, and databases. **Before:** this would export only the dashboard with ID 1, but not export associated resources.

IDs for multiple resource types can be specified:

```bash
superset-cli export --dashboard-ids=1 --chart-ids=42
```

Will export dashboard with ID 1 and related assets, as well as chart with ID 42 and related assets.

### DIFF
--- a/src/preset_cli/cli/superset/export.py
+++ b/src/preset_cli/cli/superset/export.py
@@ -77,7 +77,7 @@ def export_assets(  # pylint: disable=too-many-locals, too-many-arguments
         "chart": {int(id_) for id_ in chart_ids},
         "dashboard": {int(id_) for id_ in dashboard_ids},
     }
-    ids_requested = database_ids or dataset_ids or chart_ids or dashboard_ids
+    ids_requested = any([database_ids, dataset_ids, chart_ids, dashboard_ids])
 
     for resource_name in ["database", "dataset", "chart", "dashboard"]:
         if (not asset_types or resource_name in asset_types) and (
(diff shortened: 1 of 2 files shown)


src/preset_cli/cli/superset/export.py
```diff
@@ -77,18 +77,29 @@ def export_assets(  # pylint: disable=too-many-locals, too-many-arguments
         "chart": {int(id_) for id_ in chart_ids},
         "dashboard": {int(id_) for id_ in dashboard_ids},
     }
+    ids_requested = database_ids or dataset_ids or chart_ids or dashboard_ids
```
mistercrunch, line 80: NIT(optional): could be a tiny bit more clear is 
```python
were_any_ids_requested = any(database_ids, dataset_ids, chart_ids, dashboard_ids)
```

betodealmeida: Ah, I like it! Let me fix.

mistercrunch (approved):
Minor optional comment, otherwise LGTM!

PR #109, opened 2022-10-11
fix: prevent node from being excluded twice

betodealmeida:
The ``exclude`` option for dbt nodes was failing when a node was referenced twice; the second exclusion would fail because the node no longer existed..


src/preset_cli/cli/superset/sync/dbt/lib.py
```diff
@@ -386,20 +386,21 @@ def apply_select(
     """
     Apply dbt node selection (https://docs.getdbt.com/reference/node-selection/syntax).
     """
-    if not select:
-        return models
```
betodealmeida, line 390: Also, this was wrong. We should only return all models if `exclude` is also falsy.

PR #118, opened 2022-10-17
feat: check DB connectivity before import

betodealmeida:
Before running an import with `superset sync native` test the DB connectivity and warn if it fails (failure could be expected, depending on the network configuration).

### DIFF
--- a/src/preset_cli/cli/superset/sync/native/command.py
+++ b/src/preset_cli/cli/superset/sync/native/command.py
@@ -4,6 +4,7 @@ A command to sync Superset exports into a Superset instance.
 
 import getpass
 import importlib.util
+import logging
 import os
 from datetime import datetime, timezone
 from io import BytesIO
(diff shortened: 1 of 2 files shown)


cwegener:
Works nicely! Just noticed it when I updated to latest master today. Got slightly startled on first Warning but was happy that the import completed successfully!

betodealmeida:
> Works nicely! Just noticed it when I updated to latest master today. Got slightly startled on first Warning but was happy that the import completed successfully!

Nice! Yeah, I was expecting valid cases where this would fail as expected... but it really helps debugging in other cases.

cwegener:
One thing I forgot to mention:

The WARNING Log message does include the cleartext password in the SQLAlchemy Engine URI.

I guess that would be an easy thing to fix?

betodealmeida:
> One thing I forgot to mention:
> 
> The WARNING Log message does include the cleartext password in the SQLAlchemy Engine URI.
> 
> I guess that would be an easy thing to fix?

Oh, thanks for letting me know! That's not good, I'll fix it... I thought SQLAlchemy masked the password, I think I need to explicitly convert it to a string.

PR #122, opened 2022-10-19
fix: don't log password

betodealmeida:
Prevent logging the database password when we can't connect to it.

cc: @cwegener


cwegener:
That has resolved the issue.

PR #127, opened 2022-10-21
chore: Switching URL to api.app.preset.io

craig-rueda:
Switches manager baseUrl from `manage.app.preset.io` over to `api.app.preset.io`

### DIFF
--- a/src/preset_cli/cli/main.py
+++ b/src/preset_cli/cli/main.py
@@ -195,7 +195,7 @@ def preset_cli(  # pylint: disable=too-many-branches, too-many-locals, too-many-
 
 
 @click.command()
-@click.option("--baseurl", default="https://manage.app.preset.io/")
+@click.option("--baseurl", default="https://api.app.preset.io/")
 @click.option(
     "--overwrite",
     is_flag=True,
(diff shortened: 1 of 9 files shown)


betodealmeida:
Weird that the tests are reporting missing coverage:

```
src/preset_cli/cli/main.py                              237     24     98      1    91%   130-172
```

src/preset_cli/cli/main.py
```diff
@@ -120,7 +120,11 @@ def preset_cli(  # pylint: disable=too-many-branches, too-many-locals, too-many-
     ctx.ensure_object(dict)
 
     # store manager URL for other commands
-    ctx.obj["MANAGER_URL"] = manager_url = URL(baseurl)
+    ctx.obj["MANAGER_URL"] = manager_api_url = URL(baseurl)
+
+    if ctx.invoked_subcommand == "auth":
+        # The user is trying to auth themselves, so skip anything auth-related
+        return
```
betodealmeida, line 127: With this the credentials will never be stored (line 155) when the user runs `preset-cli auth`.

betodealmeida (approved):
Looks good, I'm just confused about the early `return` when we run `preset-cli auth`.

src/preset_cli/cli/main.py
craig-rueda: `def auth()` is invoked after this method, so that's where they will be stored. Then, the user would call something else, like `superset`, which would fetch the JWT, etc

PR #128, opened 2022-10-21
feat (dbt cloud): prompt for account/project/job

betodealmeida:
If the user doesn't specify a job ID, prompt them for one progressively (account ID, project ID, job ID).


src/preset_cli/cli/superset/sync/dbt/command.py
```diff
@@ -214,7 +302,9 @@ def dbt_cloud(  # pylint: disable=too-many-arguments, too-many-locals
     if len(databases) > 1:
         raise Exception("More than one database with the same name found")
 
-    database = databases[0]
+    # need to get the database by itself to the response has the SQLAlchemy URI
```
betodealmeida, line 305:
```suggestion
    # need to get the database by itself so the response has the SQLAlchemy URI
```

craig-rueda (approved):

Issue #130, opened 2022-10-21
Ability to sync only exposures from dbt, without syncing models


Vitor-Avila:
Users would like to be able to sync all exposures, but not all models from dbt to Preset. So, rather than determining which models exist in Preset, selecting those models, and syncing and showing exposures for those models, they would like to skip the first two steps and just show exposures.

betodealmeida:
Implemented in #132. To run it without syncing models is a bit weird, though:

```bash
$ preset-cli superset sync dbt-core ... --exposures=... --select tag:bogus
```

Where "bogus" is a tag that no model has.

PR #132, opened 2022-10-24
feat: infer dbt models from dataset

betodealmeida:
The CLI currently offers a command to populate dbt exposures — these are charts and dashboards in Superset that were created based on datasets that map back to dbt models.

Currently, the CLI uses **extra metadata that is added to datasets** when the CLI syncs them from dbt to Superset, eg:

```json
{
  "datasource_type": "table",
  "schema": "public",
  "table_name": "messages_channels",
  ...
  "extra": {
    "resource_type": "model",
    "unique_id": "model.superset_examples.messages_channels",
    "depends_on": "ref('messages_channels')"
  }
}
```

This makes the process of determining which charts and dashboards read from dbt models easy, since we can just check the extra metadata.

A customer has expressed the need for populating exposures from Preset when they have created the dataset manually. Since the datasets don't have the extra metadata, we need to infer the model references from the schema and table name.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,5 +1,5 @@
 [flake8]
-ignore = E203, E266, E501, W503, F403, F401
+ignore = E203, E266, E501, W503, F403, F401, W293
 max-line-length = 79
 max-complexity = 18
 select = B,C,E,F,W,T4,B9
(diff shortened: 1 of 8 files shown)


betodealmeida:
Closes https://github.com/preset-io/backend-sdk/issues/130.

eschutho (approved):

PR #134, opened 2022-10-25
feat: refresh JWT whenever needed

betodealmeida:
A customer needs to run a really big import (~12 hours), but the JWT is only valid for 5. So I added some logic to refresh authentication (across all mechanisms) whenever needed.

The main auth class (simplified) now looks like this:

```python
class Auth:  # pylint: disable=too-few-public-methods
    def __init__(self):
        self.session = Session()

        # call ``self.reauth`` after every request
        self.session.hooks["response"].append(self.reauth)

    def auth(self) -> None:
        raise NotImplementedError("Must be implemented for reauthorizing")

    def reauth(self, r: Response, *args: Any, **kwargs: Any) -> Response:
        if r.status_code != 401:
            return r

        # if the response was 401, auth and try again
        try:
            self.auth()
        except NotImplementedError:
            return r

        self.session.headers.update(self.get_headers())
        r.request.headers.update(self.get_headers())
        return self.session.send(r.request, verify=False)
```

On `__init__` it attaches a hook to the session, calling `reauth` **after** every request. On `reauth`, if the response was a 401 and the derived class implements the `auth` method, then `auth` gets called to refresh cookies/tokens, and the request is resent.

This PR implements `auth` for the Preset and the username:password authentication mechanisms.

### DIFF
--- a/src/preset_cli/api/clients/dbt.py
+++ b/src/preset_cli/api/clients/dbt.py
@@ -581,7 +581,7 @@ class DBTClient:  # pylint: disable=too-few-public-methods
         self.graphql_client = GraphqlClient(endpoint=GRAPHQL_ENDPOINT)
         self.baseurl = REST_ENDPOINT
 
-        self.session = auth.get_session()
+        self.session = auth.session
         self.session.headers.update(auth.get_headers())
         self.session.headers["User-Agent"] = "Preset CLI"
         self.session.headers["X-Client-Version"] = __version__
--- a/tests/auth/preset_test.py
+++ b/tests/auth/preset_test.py
@@ -1,0 +1,89 @@
+"""
+Test Preset auth.
+"""
+
+import pytest
+from pytest_mock import MockerFixture
+from requests_mock.mocker import Mocker
+from yarl import URL
+
+from preset_cli.auth.preset import PresetAuth
+
+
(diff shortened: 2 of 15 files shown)


src/preset_cli/api/clients/dbt.py
```diff
@@ -581,7 +581,7 @@ def __init__(self, auth: Auth):
         self.graphql_client = GraphqlClient(endpoint=GRAPHQL_ENDPOINT)
         self.baseurl = REST_ENDPOINT
 
-        self.session = auth.get_session()
+        self.session = auth.session
```
betodealmeida, line 584: I got rid of `get_session`, since it only added noise.

tests/auth/preset_test.py
```diff
@@ -0,0 +1,89 @@
...
+        status_code=401,
+    )
+    requests_mock.get(
+        "https://api.app.preset.io/",
+        request_headers={"Authorization": "Bearer JWT_TOKEN2"},
+        status_code=200,
+    )
+
+    auth = PresetAuth(URL("https:/api.app.preset.io/"), "TOKEN", "SECRET")
+    print(auth.get_headers())
```
betodealmeida, line 47:
```suggestion
```

craig-rueda (approved):

PR #135, opened 2022-10-26
feat: retry imports on connection error

betodealmeida:
A customer is running a big import, and mid-process the CLI will fail with a `ConnectionError`. I added retries to the import command trying to fix it.

### DIFF
--- a/src/preset_cli/cli/superset/sync/native/command.py
+++ b/src/preset_cli/cli/superset/sync/native/command.py
@@ -6,7 +6,7 @@ import getpass
 import importlib.util
 import logging
 import os
-from datetime import datetime, timezone
+from datetime import datetime, timedelta, timezone
 from io import BytesIO
 from pathlib import Path
 from types import ModuleType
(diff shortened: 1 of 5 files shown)


src/preset_cli/cli/superset/sync/native/command.py
```diff
@@ -213,14 +215,35 @@ def import_resources_individually(
...
         for path, config in configs.items():
-            if path.parts[1] == resource_name:
-                asset_configs = {path: config}
-                for uuid in get_related_uuids(config):
-                    asset_configs.update(related_configs[uuid])
-                _logger.info("Importing %s", path.relative_to("bundle"))
-                contents = {str(k): yaml.dump(v) for k, v in asset_configs.items()}
-                import_resources(contents, client, overwrite)
-                related_configs[config["uuid"]] = asset_configs
+            if path.parts[1] != resource_name:
```
craig-rueda, line 218: Would be bool to use a backoff here :).

https://pypi.org/project/backoff/

craig-rueda (approved):

PR #136, opened 2022-10-26
chore: clean `get_resource`

betodealmeida:
`SupersetClient` has 2 methods for retrieving assets from Superset:

- `get_resources`, which returns a list of resources (eg, charts).
- `get_resource`, which returns a single resource given its ID and type.

Their response is slightly different. The former returns a subset of the JSON response from Superset, containing only the actual asset (it returns `response_payload["result"]`). The latter returns the full JSON response from Superset, so after the code has to call:

```python
dataset = client.get_dataset(1)["result"]
```
 
I cleaned up the client so that only the result is returned, allowing code to be rewritten as:

```python
dataset = client.get_dataset(1)
```

### DIFF
--- a/src/preset_cli/api/clients/superset.py
+++ b/src/preset_cli/api/clients/superset.py
@@ -296,7 +296,7 @@ class SupersetClient:  # pylint: disable=too-many-public-methods
         """
         Run a dimensional query.
         """
-        dataset = self.get_dataset(dataset_id)["result"]
+        dataset = self.get_dataset(dataset_id)
 
         if time_column is None:
             time_columns = [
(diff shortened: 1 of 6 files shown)


src/preset_cli/api/clients/superset.py
```diff
@@ -394,7 +394,7 @@ def get_resource(self, resource_name: str, resource_id: int) -> Any:
         response = self.session.get(url)
         validate_response(response)
 
-        resource = response.json()
+        resource = response.json()["result"]
```
hughhhh, line 397: Are we okay with throwing a KeyError if result isn't available?

betodealmeida: Yeah, if the key is not there it's because the Superset API has completely changed, so it's OK. Also, today we'd also raise `KeyError`, only later in the code.

AAfghahi (approved):

PR #142, opened 2022-11-03
feat (dbt): allow syncing only exposures

betodealmeida:
Allow syncing only exposures from Preset workspaces to a YAML file.

This now works with both dbt Cloud and dbt Core:

```bash
$ preset-cli superset sync dbt-core ... --exposures=exposures.yaml --exposures-only
$ preset-cli superset sync dbt-cloud ... --exposures=exposures.yaml --exposures-only
```

The datasets don't have to be created via the CLI for the exposure discovery to work. The command will look at all models (filtered by `--select` and `--excluded`, if set) and figure out which datasets correspond to them based on the the database, schema, and table name.

### DIFF
--- a/src/preset_cli/cli/superset/sync/dbt/command.py
+++ b/src/preset_cli/cli/superset/sync/dbt/command.py
@@ -18,7 +18,7 @@ from preset_cli.api.clients.superset import SupersetClient
 from preset_cli.auth.token import TokenAuth
 from preset_cli.cli.superset.sync.dbt.databases import sync_database
 from preset_cli.cli.superset.sync.dbt.datasets import sync_datasets
-from preset_cli.cli.superset.sync.dbt.exposures import sync_exposures
+from preset_cli.cli.superset.sync.dbt.exposures import ModelKey, sync_exposures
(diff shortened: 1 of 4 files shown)


betodealmeida:
This closes #142

anupreet (approved):
based on your tests and dbt documentation, the code LGTM

src/preset_cli/cli/superset/sync/dbt/command.py
```diff
@@ -75,6 +81,7 @@ def dbt_core(  # pylint: disable=too-many-arguments, too-many-locals
     import_db: bool = False,
     disallow_edits: bool = True,
     external_url_prefix: str = "",
+    exposures_only: bool = True,
```
eschutho, line 84: since you have this defaulting to False above, do you want the default here to match?

betodealmeida: Oh, good catch!

betodealmeida: It doesn't really matter, but it's good to be consistent. Updated it.

PR #145, opened 2022-11-07
updating sync_database

crysensible:
dbt_project.yml specifies the profile to use in profiles.yml. 

The project_name and profile_name in dbt_project.yml can be different.

The reference between dbt_project.yml to use the right profile in profiles.yml should reference dbt_project[profile].


crysensible:
@betodealmeida - I'm a first time contributor. Would you be able to look at this or advise since the build workflows are not running? Thank you in advance!

Issue #144, opened 2022-11-07
sync_database doesn't find existing database 


crysensible:
In the Preset UI, the database connections front end, the rule for the name of a database is 'Copy the name of the database you are trying to connect to.'  Because of that, the database where the final models are stored in Snowflake is the default name of the database connection in Preset. For example, in Snowflake, my database that I read my models off of is called 'ANALYTICS_PROD'

In sync_database, it seems that the database_name being searched for is: 
database_name = meta.pop("database_name", f"{project_name}_{target_name}")

When I run dbt-core, it's not finding the database and I am curious if it's looking for that database_name. I think it should be looking for --> dbt_project --> profile --> target --> outputs --> database

Vitor-Avila:
Hi @crysensible, 

Just to make sure that we're on the same page: You would like the CLI to actually use the db name available under ` dbt_project --> profile --> target --> outputs --> database`, instead of the current logic?

One thing that could be used to properly map the databases, is to specify the Preset DB name on your `Profiles.yml` file, as indicated [here](https://github.com/preset-io/backend-sdk#synchronizing-to-and-from-dbt):

![image](https://user-images.githubusercontent.com/96086495/201177160-373a7b34-bf85-4140-9ac7-54dc7d7e1574.png)

Let me know if that would solve this issue.

crysensible:
Hi Vitor, thank you for getting back to me. 

I have added it to my profiles.yml file and tried to run the command again but it still fails to find the database.

Here's my setup in profiles.yml
<img width="185" alt="image" src="https://user-images.githubusercontent.com/110119242/201447996-235ba482-ac11-4dd0-bc25-c056713f5fc2.png">

Here is the command I am running:
`preset-cli --workspaces=my_workspace superset sync dbt-core ~/dbt_project.yml --project=snowflake --target=prod --profiles=~/.dbt/profiles.yml --select=model_name`

This is what's returned:
"No database was found, pass --import-db to create"

What do you suggest?

Vitor-Avila:
@crysensible I know Juliann is in contact with you to discuss and review the process to contribute to the CLI. In the meantime, we're working to push your suggested change on this PR: https://github.com/preset-io/backend-sdk/pull/151

crysensible:
Perfect, thank you @Vitor-Avila and @eschutho.

crysensible:
Hi team, is there any update on this?

eschutho:
[...]

PR #151, opened 2022-11-17
use profile instead of project on db lookup

eschutho:
Fixes a bug where the db lookup was keying off the project name instead of the profile name when syncing a  database. 

I changed the profile name in the tests to be different from the project so that we can double-check where the lookup is happening. 

fixes: https://github.com/preset-io/backend-sdk/issues/144


betodealmeida (approved):
This makes sense... thanks for fixing it, Elizabeth! I'll check with some of our customers if it works for them before making a new release, just to be safe.

Issue #126, opened 2022-10-21
Support dbt-core 1.3


cwegener:
dbt-core 1.3 changes the metric spec as follows:

- `type` --> `calculation_method`
- `sql` --> `expression`
- `type: expression` --> `calculation_method: derived`

`superset-cli` should:

1. Either know which version of dbt-core is used and use the correct marshmallow MetricSchema version depending on version number
2. Or dynamically infer the MetricSchema during runtime (with `Schema.from_dict`)

From looking at dbt-cloud API, there seems to be know way to dynamically detect the dbt-core version of a particular dbt-cloud environment.

betodealmeida:
Ah, thanks @cwegener! I'll fix this!

cwegener:
I'm sure there are likely more options of how to do this in marshmallow. Actually on second thought, dynamically inferring the schema alone is not sufficient to deal with this.

PaxonF:
Any chance support for dbt-core 1.3 will be added soon? I'm looking to sync metrics to Preset, but I do run into some key errors when trying to create a metric with `calculation_method: average`.

`File "/Users/paxonfischer/Documents/GitHub/dbt/.venv/lib/python3.8/site-packages/preset_cli/cli/superset/sync/dbt/metrics.py", line 35, in get_metric_expression
    type_ = metric["type"]`

cwegener:
Same issue here. Just doing my testing of #154 and I do get the `KeyError` as well.

betodealmeida:
This should be fixed in #159.

PR #154, opened 2022-11-30
fix (dbt): derived metrics

### LINKED ISSUE #153: Expression metrics are not imported to Superset
### Description
Users are able to specify metrics on the `schema.yml` file, that should be created on Superset as dataset metrics. However, if the metric is an `expression`, since they don't have a `model` key, they don't get associated to any dataset on Superset.

### How to reproduce the issue
1. Create expression metrics on your `schema.yml` file. For example:
```
metrics:
  - name: total_revenue
    label: SUM of all price_each
    model: ref('modeled_sales')
    description: "The SUM of all price_each"
    type: sum
    sql: price_each
    timestamp: date
    time_grains: [day, week, month, year]
  - name: revenue_multiplied
    label: "Revenue calculated differently"
    description: "Let's try to use an expression"
    type: expression
    sql: "{{metric('total_revenue')}} * 1.25"
    timestamp: date
    time_grains: [day, week, month, year]
```
2. Sync data to Superset.

### Expected Results
Both `total_revenue` and `revenue_multiplied` should be created on the `modeled_sales` dataset in Superset.

### Actual Results
Only `total_revenue` is created on the `modeled_sales` dataset in Superset.

betodealmeida:
Some improvements to derived metrics in dbt:

- Derived metrics were not being synced because we were not mapping them to the correct models. This is now fixed.
- dbt 1.3 introduced changes in the schema of metrics. Both schemas are now supported.

Closes #153 and #126.


src/preset_cli/cli/superset/sync/dbt/command.py
```diff
@@ -150,7 +150,7 @@ def dbt_core(  # pylint: disable=too-many-arguments, too-many-locals
         for config in configs["metrics"].values():
             # conform to the same schema that dbt Cloud uses for metrics
             config["dependsOn"] = config["depends_on"]["nodes"]
-            config["uniqueID"] = config["unique_id"]
+            config["uniqueId"] = config["unique_id"]
```
betodealmeida, line 153: This was a bug, but since the metric unique ID was not needed anywhere it was not caught. With the changes in this PR we use the metric ID to traverse the DAG upstream to find parent models.

src/preset_cli/cli/superset/sync/dbt/datasets.py
```diff
@@ -109,9 +112,7 @@ def sync_datasets(  # pylint: disable=too-many-locals, too-many-branches, too-ma
 
         dataset_metrics = []
         model_metrics = {
-            metric["name"]: metric
-            for metric in metrics
-            if model["unique_id"] in metric["depends_on"]
```
betodealmeida, line 114: This was too naive, and didn't account for how derived metrics **depend on other metrics**, instead of depending on models.

src/preset_cli/cli/superset/sync/dbt/metrics.py
```diff
@@ -59,3 +70,47 @@ def apply_filters(sql: str, filters: List[FilterSchema]) -> str:
...
+    related_metrics = []
+
+    for metric in metrics:
+        parents = set()
+        queue = [metric]
+        while queue:
+            node = queue.pop()
+            depends_on = node["depends_on"]
+            if is_derived(node):
+                queue.extend(metric_map[parent] for parent in depends_on)
```
betodealmeida, line 102: This is the crux of the PR: for derived metrics we need to look at the upstream metrics to find the parent models.

stevepisani: is the metric list something from superset or something from the dbt manifest?

src/preset_cli/cli/superset/sync/dbt/metrics.py
```diff
@@ -22,8 +25,16 @@ def get_metric_expression(metric_name: str, metrics: Dict[str, MetricSchema]) ->
         raise Exception(f"Invalid metric {metric_name}")
 
     metric = metrics[metric_name]
-    type_ = metric["type"]
-    sql = metric["sql"]
+    if "calculation_method" in metric:
```
hughhhh, line 28: nit: have it inline
```
legacy = "calculation_method" in metric
type_ = metric["calculation_method"]  if legacy else metric["type"]
sql = metric["expression"] if legacy else metric["sql"]
expression = "derived" if legacy else  "expression"
```

stevepisani (approved):
looks good to me!

hughhhh (approved):
LGTM

cwegener:
Exciting work!! Can't wait to make some time and test this out.

cwegener:
@betodealmeida The `MetricSchema` marshmallow schema is missing the new dbt 1.3 field names, so the new fields names never actually get passed through to this new dbt code.

betodealmeida:
Ah, thanks! Let me fix it.

PR #162, opened 2022-12-07
fix:Minor changes

Vitor-Avila:
* Fixed two typos
* Removed the `--overwrite` flag from the dbt sync since it's not required (that's controlled by the `--import-db`).


betodealmeida (approved):
Thanks!

PR #169, opened 2022-12-13
Fixes #168 and #177

### LINKED ISSUE #168: The import-db flag is not functioning as expected
### Current Problem:
If you run a dbt sync with the preset-cli, regardless of setting the `--import-db` flag, if a database is found in Preset, it will overwrite the connect.  The problem is that the database user used during sync may not be the same user that will be used in Preset.  

## Solution:
Make it so if the `--import-db` setting is not present, skip overwriting the database connection.

dweaver33:
@betodealmeida, please check over to see if this fixes the issue I opened.


dweaver33:
Tested locally and behaves as I would expect for my purpose.  

Behavior is as follows:
* `--import-db` is NOT set
  * If DB exists, then nothing happens
  * If DB does not exist, then DB is created
* `--import-db` is set
  * If DB exists, it will overwrite with new credentials
  * If DB does not exist, then DB is created

dweaver33:
Also added a fix for #177

Issue #170, opened 2022-12-14
Make import-roles idempotent


Vitor-Avila:
### Current Scenario
Currently, the `--import-roles` operation won't properly replace existing DARs on the destination Workspace.

### Suggested improvement
Make the `--import-roles` operation idempotent, so that it can be used for continuous sync.

Vitor-Avila:
Implemented on https://github.com/preset-io/backend-sdk/pull/179.

PR #179, opened 2023-01-17
feat: make import-roles idempotent

betodealmeida:
Make `import-roles` update the role if it already exists.


betodealmeida:
Closes https://github.com/preset-io/backend-sdk/issues/170.

hughhhh (approved):

PR #182, opened 2023-01-24
update dbt sync section on readme

Vitor-Avila:
* Moved the DB creation/link part to the dbt core section
* Added a section to highlight that the db connection must already exist on the workspace when syncing from dbt cloud, and that the display name must match the db name.


betodealmeida (approved):
Thanks!

Issue #177, opened 2022-12-30
Exposure names contain invalid characters


dweaver33:
Currently when sync'ing with dbt, the exposures from Preset use the `name` property to pull the name of the exposure.  This causes problems as the name defined in Preset can contain invalid characters for dbt exposure names.  Per [documentation here](https://docs.getdbt.com/reference/exposure-properties), exposure names can only contain letters, numbers, and underscores (no spaces or special characters)

I propose we change the behavior to take the Preset name, cleanse it to contain only letters, numbers, and underscores (no spaces or special characters) and use this for the `name` property. 

Then we can use the original Preset name in the `label` property.

Vitor-Avila:
Hey @dweaver33, 

Sorry for the delay here. I'm working on this issue this week, and should have a PR soon.

dweaver33:
Thank you!

Vitor-Avila:
I have a potential fix here: https://github.com/preset-io/backend-sdk/pull/221

dweaver33:
Thank you!  We'll do some testing and let you know.

PR #193, opened 2023-03-03
(chore) Supporting dbt sync without updating the db connection configuration on Superset

Vitor-Avila:
Currently, the CLI would always update the DB connection configuration on the Workspace using the information from the `profiles.yml` file (regardless if `--import-db` was passed or not). The flag was only validated in case **no database connection was found**.

This PR introduces the ability to perform a dbt sync without updating the DB connection. This is very useful in case the dbt configuration uses different credentials, or even a different authentication method that might not be supported by Superset.

### DIFF
--- a/src/preset_cli/cli/superset/sync/dbt/databases.py
+++ b/src/preset_cli/cli/superset/sync/dbt/databases.py
@@ -53,11 +53,11 @@ def sync_database(  # pylint: disable=too-many-locals, too-many-arguments
     if base_url and "external_url" not in meta:
         meta["external_url"] = str(base_url.with_fragment("!/overview"))
 
-    if databases:
-        _logger.info("Found an existing database, updating it")
+    if import_db and databases:
+        _logger.info("Found an existing database connection, updating it")
         database = databases[0]
-
         meta.pop("uuid", None)
+
         database = client.update_database(
             database_id=database["id"],
             database_name=database_name,
@@ -66,10 +66,12 @@ def sync_database(  # pylint: disable=too-many-locals, too-many-arguments
             sqlalchemy_uri=connection_params["sqlalchemy_uri"],
             **meta,
         )
-    elif not import_db:
-        raise DatabaseNotFoundError()
-    else:
-        _logger.info("No database found, creating it")
+
+        database["sqlalchemy_uri"] = connection_params["sqlalchemy_uri"]
+
+    elif import_db:
+        _logger.info("No database connection found, creating it")
+
         database = client.create_database(
             database_name=database_name,
             is_managed_externally=disallow_edits,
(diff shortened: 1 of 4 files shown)


src/preset_cli/cli/superset/sync/dbt/databases.py
```diff
@@ -78,6 +80,16 @@ def sync_database(  # pylint: disable=too-many-locals, too-many-arguments
             **meta,
         )
 
-    database["sqlalchemy_uri"] = connection_params["sqlalchemy_uri"]
```
Vitor-Avila, line 81: Previously, this would always be executed (which pretty much builds the `SQLAlchemy` URI from the connection settings on the `profiles.yml`). Now it's only executed in case `--import-db` is passed.

src/preset_cli/cli/superset/sync/dbt/databases.py
```diff
@@ -78,6 +80,16 @@ def sync_database(  # pylint: disable=too-many-locals, too-many-arguments
...
+        database["sqlalchemy_uri"] = connection_params["sqlalchemy_uri"]
+
+    elif databases:
+        _logger.info("Found an existing database connection, using it")
+        database = databases[0]
+        database["sqlalchemy_uri"] = client.get_database(database["id"])[
```
Vitor-Avila, line 88: In case `--import-db` is not passed, I added this to fetch the `SQLAlchemy` URI from the Superset API instead. The password wouldn't be included, but I don't think this would be a problem since we're not updating the DB connection.

dweaver33:
I've covered this fix here:  https://github.com/preset-io/backend-sdk/pull/169

Vitor-Avila:
Hey @dweaver33, 

Thank you so much for your interest on contributing to the CLI. I'm afraid that due to license restrictions, we're currently not accepting contributions from outside of Preset. I was actually going to add a comment to your PR/Issue once this gets merged.

Thank you so much for your understanding.

dweaver33:
> Hey @dweaver33,
> 
> Thank you so much for your interest on contributing to the CLI. I'm afraid that due to license restrictions, we're currently not accepting contributions from outside of Preset. I was actually going to add a comment to your PR/Issue once this gets merged.
> 
> Thank you so much for your understanding.

Ha!  Didn't realize you were a maintainer here. 

In that case, can you also look at my fix in that same PR for the exposures?  #177

betodealmeida (approved):
This looks great! Thanks, Vitor!

Vitor-Avila:
> > Hey @dweaver33,
> > Thank you so much for your interest on contributing to the CLI. I'm afraid that due to license restrictions, we're currently not accepting contributions from outside of Preset. I was actually going to add a comment to your PR/Issue once this gets merged.
> > Thank you so much for your understanding.
> 
> Ha! Didn't realize you were a maintainer here.
> 
> In that case, can you also look at my fix in that same PR for the exposures? #177

I'll take a look on that ticket, @dweaver33. Thank you so much!

PR #200, opened 2023-04-03
Do not build SQLAlchemy URI in case the database won't be imported/modified

Vitor-Avila:
Currently, for the dbt sync the database connection would only be created/updated in Superset in case `--import-db` was passed. However, the CLI was still building the SQLAlchemy URI regardless of this flag, which would still throw an error in case the DB authentication method in DBT is un-supported in Preset. 

With this PR, `build_sqlalchemy_params()` is only called in case `import_db == True`.

### DIFF
--- a/src/preset_cli/cli/superset/sync/dbt/databases.py
+++ b/src/preset_cli/cli/superset/sync/dbt/databases.py
@@ -49,10 +49,10 @@ def sync_database(  # pylint: disable=too-many-locals, too-many-arguments
         meta["external_url"] = str(base_url.with_fragment("!/overview"))
 
     if import_db:
-        if "connection_params" in meta:
-            connection_params = meta.pop("connection_params")
-        else:
-            connection_params = build_sqlalchemy_params(target)
+        connection_params = meta.pop(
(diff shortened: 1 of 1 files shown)


src/preset_cli/cli/superset/sync/dbt/databases.py
```diff
@@ -53,32 +48,36 @@ def sync_database(  # pylint: disable=too-many-locals, too-many-arguments
...
-            is_managed_externally=disallow_edits,
-            masked_encrypted_extra=connection_params.get("encrypted_extra"),
-            **connection_params,
-            **meta,
-        )
+    if import_db:
+        if "connection_params" in meta:
+            connection_params = meta.pop("connection_params")
+        else:
+            connection_params = build_sqlalchemy_params(target)
```
betodealmeida, line 55: Just realized this can be done in a single line:
```suggestion
        connection_params = meta.pop(
            "connection_params",
            build_sqlalchemy_params(target),
        )
```

betodealmeida (approved):

PR #203, opened 2023-05-03
feat: allow passing a certification payload

betodealmeida:
Allow passing a certification payload when syncing datasets from dbt.

### DIFF
--- a/src/preset_cli/cli/superset/sync/dbt/datasets.py
+++ b/src/preset_cli/cli/superset/sync/dbt/datasets.py
@@ -73,6 +73,7 @@ def sync_datasets(  # pylint: disable=too-many-locals, too-many-branches, too-ma
     database: Any,
     disallow_edits: bool,
     external_url_prefix: str,
+    certification=None,
 ) -> List[Any]:
     """
     Read the dbt manifest and import models as datasets with metrics.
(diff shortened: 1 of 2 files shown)


src/preset_cli/cli/superset/sync/dbt/datasets.py
```diff
@@ -73,6 +73,7 @@ def sync_datasets(  # pylint: disable=too-many-locals, too-many-branches, too-ma
     database: Any,
     disallow_edits: bool,
     external_url_prefix: str,
+    certification=None,
```
zephyring, line 76: can we specify the type hint here

tests/cli/superset/sync/dbt/datasets_test.py
```diff
@@ -275,6 +275,65 @@ def test_sync_datasets_no_metrics(mocker: MockerFixture) -> None:
...
+    }
+
+    sync_datasets(
+        client=client,
+        models=models,
+        metrics=[],
+        database={"id": 1, "sqlalchemy_uri": "postgresql://user@host/examples_dev"},
+        disallow_edits=False,
+        external_url_prefix="",
+        certification={"details": "This dataset is synced from dbt Cloud"},
```
zephyring, line 296: do we already have another test case when cert is None

betodealmeida: We do!

src/preset_cli/cli/superset/sync/dbt/datasets.py
betodealmeida: Ah, I forgot! Sorry, will fix!

zephyring (approved):

Issue #187, opened 2023-02-09
Ability to filter for dbt models based on the materialized field


Vitor-Avila:
Currently, the `--exclude` flag only allows to match a `tag`. It would be good to extend the `--exclude` flag to also include the materialized field in the model config.

Vitor-Avila:
For example, users can run on dbt:
```
dbt run --exclude config.materialized:false
```

[Reference](https://docs.getdbt.com/reference/node-selection/methods#the-config-method).

PR #204, opened 2023-05-05
Adding config option for filtering models

hughhhh:
Users now will be able to filter `Model.config` values when running the cli
```
$ dbt run --select config.materialized:view

# w/ config
{{ config(
  materialized = 'view',
  unique_key = ['column_a', 'column_b'],
  grants = {'select': ['reporter', 'analysts']},
  transient = true
) }}
```

### DIFF
--- a/tests/cli/superset/sync/dbt/lib_test.py
+++ b/tests/cli/superset/sync/dbt/lib_test.py
@@ -391,7 +391,7 @@ def test_filter_models() -> None:
             "persist_docs": {},
             "quoting": {},
             "column_types": {},
-        }
+        },
     }
     models: List[ModelSchema] = [one, two, three]  # type: ignore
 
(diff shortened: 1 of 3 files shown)


hughhhh:
resolves https://github.com/preset-io/backend-sdk/issues/187

tests/cli/superset/sync/dbt/lib_test.py
```diff
@@ -386,6 +386,12 @@ def test_filter_models() -> None:
         "unique_id": "model.three",
         "depends_on": ["source.zero"],
         "children": ["model.two"],
+        "config": {
```
betodealmeida, line 389: Looks like the tests are failing because you need to add `config` to the other models, even if it's an empty dict.

betodealmeida (approved):
Awesome!

Issue #188, opened 2023-02-15
The CLI is not escaping Jinja logic syntax, only macros during export


Vitor-Avila:
## Description

When exporting an asset via the CLI that contains a Jinja macro, the CLI would escape the syntax, since it shouldn't be considered (by the CLI) during import - it's only used by Superset. As a consequence, this syntax:
```
  and country in {{ filter_values('country')|where_in }}
``` 
Becomes this:
```
and country in {{ '{{' }} filter_values('country')|where_in {{ '}}' }}
```

However, logic statements aren't being escaped, which causes the import operation to fail.

## Solution 
The CLI should also escape logic statements. For example, below block:
```
{% if filter_values('country')|length %}
  and country in {{ filter_values('country')|where_in }}
{% endif %}
```

should become:
```
{{ '{% if' }} filter_values('country')|length {{ '%}' }}
  and country in {{ '{{' }} filter_values('country')|where_in {{ '}}' }}
{{ '{% endif %}' }}
```

Vitor-Avila:
Closed with https://github.com/preset-io/backend-sdk/pull/205

PR #205, opened 2023-05-09
fix: Escaping Jinja logical statements from assets

Vitor-Avila:
This PR implements escaping for Jinja logical statements (https://github.com/preset-io/backend-sdk/issues/188), and also adds two new flags:
* `--disable-jinja-templating`: This flag was added to the `sync native` command, so that the CLI doesn't render any Jinja templating from the assets. This is useful in case the assets contain Jinja and were exported via the UI (and therefore the syntax wasn't escaped).
* `--disable-jinja-escaping`: This flag was added to the `export-assets` command, so that the CLI doesn't escape Jinja syntax from the YAML files. This is useful in case the assets contain Jinja and would be imported through the UI.

Also, these two flags can be used in case the escaping/rendering logic is incorrectly escaping/rendering incorrect values (like JSON curly brackets, etc).

### DIFF
--- a/src/preset_cli/cli/superset/sync/native/command.py
+++ b/src/preset_cli/cli/superset/sync/native/command.py
@@ -146,7 +146,7 @@ def render_yaml(path: Path, env: Dict[str, Any]) -> Dict[str, Any]:
     help="Split imports into individual assets",
 )
 @click.pass_context
-def native(  # pylint: disable=too-many-locals, too-many-arguments
+def native(  # pylint: disable=too-many-locals, too-many-arguments, too-many-branches
     ctx: click.core.Context,
     directory: str,
     option: Tuple[str, ...],
--- a/src/preset_cli/cli/superset/sync/native/command.py
+++ b/src/preset_cli/cli/superset/sync/native/command.py
@@ -80,6 +80,16 @@ def is_yaml_config(path: Path) -> bool:
     )
 
 
+def load_yaml(path: Path) -> Dict[str, Any]:
+    """
+    Load a YAML file and returns it as a dictionary.
+    """
+    with open(path, encoding="utf-8") as input_:
+        content = input_.read()
+
+    return yaml.load(content, Loader=yaml.SafeLoader)
+
+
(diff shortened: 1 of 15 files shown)


src/preset_cli/cli/superset/sync/native/command.py
```diff
@@ -171,11 +188,17 @@ def native(  # pylint: disable=too-many-locals, too-many-arguments
         if path_name.is_dir() and not path_name.stem.startswith("."):
             queue.extend(path_name.glob("*"))
         elif is_yaml_config(relative_path):
-            config = render_yaml(path_name, env)
+            if disable_jinja_templating:
+                config = load_yaml(path_name)
+            else:
+                config = render_yaml(path_name, env)
```
betodealmeida, line 194: Small nit:

```python
config = load_yaml(path_name) if disable_jinja_templating else render_yaml(path_name, env)
```

src/preset_cli/cli/superset/sync/native/command.py
```diff
@@ -171,11 +188,17 @@ def native(  # pylint: disable=too-many-locals, too-many-arguments
...
 
             overrides_path = path_name.with_suffix(".overrides" + path_name.suffix)
             if overrides_path.exists():
-                overrides = render_yaml(overrides_path, env)
+                if disable_jinja_templating:
+                    overrides = load_yaml(overrides_path)
+                else:
+                    overrides = render_yaml(overrides_path, env)
```
betodealmeida, line 201: Same here.

betodealmeida (approved):
sagan

src/preset_cli/cli/superset/sync/native/command.py
Vitor-Avila: thanks! changed as suggested

src/preset_cli/cli/superset/sync/native/command.py
Vitor-Avila: changed this one as well.

PR #211, opened 2023-05-30
fix(superset): `get_database`

betodealmeida:
Due to a CVE, some of the information available in `api/v1/database/$id` was moved to `api/v1/database/$id/connection`. This PR updates the Superset client to try both endpoints.

Added a test covering the case.

### DIFF
--- a/src/preset_cli/api/clients/superset.py
+++ b/src/preset_cli/api/clients/superset.py
@@ -477,7 +477,17 @@ class SupersetClient:  # pylint: disable=too-many-public-methods
         """
         Return a single database.
         """
-        return self.get_resource("database", database_id)
+        database = self.get_resource("database", database_id)
+        if "sqlalchemy_uri" in database:
+            return database
+
+        url = self.baseurl / "api/v1/database" / str(database_id) / "connection"
+        response = self.session.get(url)
+        validate_response(response)
+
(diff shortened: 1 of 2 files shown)


src/preset_cli/api/clients/superset.py
```diff
@@ -477,7 +477,17 @@ def get_database(self, database_id: int) -> Any:
...
-        return self.get_resource("database", database_id)
+        database = self.get_resource("database", database_id)
+        if "sqlalchemy_uri" in database:
+            return database
+
+        url = self.baseurl / "api/v1/database" / str(database_id) / "connection"
+        response = self.session.get(url)
+        validate_response(response)
+
+        resource = response.json()["result"]
```
zephyring, line 488: it will be nice to have a schema for the return object

zephyring (approved):

src/preset_cli/api/clients/superset.py
zephyring: I am not familiar with the above 2 endpoints but another option is to introduce a new method for the connection endpoint.
The API client should be as dumb as possible.

betodealmeida: Good point, I have a TODO in that file to add them. It's not super trivial because the schema depends on the version of Superset running.

PR #213, opened 2023-05-31
feat: sync certification information from dbt models

Vitor-Avila:
This PR introduces the ability to control `certification` information from the dbt models definition. Previously, the certification would be automatically set to:
``` json
{
    "certification":
    {
        "details": "This table is produced by dbt"
    }
```

Now, the `certification` can be defined on the `meta.superset.extra.certification` field. For example:

``` yaml
models:
  - name: dbt_vehicle_sales
    meta:
      superset:
        extra:
          certification: 
            certified_by: Data Analytics team
            details: Vehicle Sales data
```

It's also possible to set `certification` to `null`, to prevent displaying the pop up in the Superset UI (useful in case the model is not ready yet for Production use):

``` yaml
models:
  - name: dbt_vehicle_sales
    meta:
      superset:
        extra:
          certification: null
```

Additional fields can also be specified for the `extra` field, such as the `warning_markdown`:

``` yaml
models:
  - name: dbt_vehicle_sales
    meta:
      superset:
        extra:
          warning_markdown: UNDER CONSTRUCTION
```

Added some new test scenarios for the dataset sync.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,11 @@
 Changelog
 =========
 
+Version 0.2.2 - TBD
+==========================
+
+- Added the ability to sync ``certification`` and additional ``extra`` information from dbt models (`# <>`_).
+
 Version 0.2.1 - 2023-05-30
 ==========================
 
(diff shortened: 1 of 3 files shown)


betodealmeida (approved):
tenor-1679277363

PR #215, opened 2023-06-05
fix: Removing the certification key in case certification is null

Vitor-Avila:
This PR adds a few fixes/improvements to the dataset certification flow:
* In case the `certification` is set to `null` via the dbt model definition, before we were setting `certification: null` in the dataset, which wouldn't show the certification pop up, but it would still include the dataset when filtering for `certified: yes`. With these changes, in case `certification` is set to `null` in the model definition, the `certification` key won't be included in the dataset's `extra` field.
* Renamed the `certification` local variable to `certification_info` to better distinguish from the function argument. Also changed its content to a dictionary to only include the key pair in the `extra` definition in case it contains an actual value.
* Updated the test for the new desired flow.

### DIFF
--- a/src/preset_cli/cli/superset/sync/dbt/datasets.py
+++ b/src/preset_cli/cli/superset/sync/dbt/datasets.py
@@ -86,11 +86,13 @@ def sync_datasets(  # pylint: disable=too-many-locals, too-many-branches, too-ma
 
         # load additional metadata from dbt model definition
         model_kwargs = model.get("meta", {}).pop("superset", {})
-        certification = (
-            model_kwargs.get("extra", {}).pop("certification")
-            if "certification" in model_kwargs
-            else certification or {"details": "This table is produced by dbt"}
-        )
+        certification_info = {
+            "certification": (
+                model_kwargs.get("extra", {}).pop("certification")
+                if "certification" in model_kwargs.get("extra", {})
+                else certification or {"details": "This table is produced by dbt"}
+            ),
+        }
 
         filters = {
             "database": OneToMany(database["id"]),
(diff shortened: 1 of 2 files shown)


src/preset_cli/cli/superset/sync/dbt/datasets.py
```diff
@@ -86,11 +86,13 @@ def sync_datasets(  # pylint: disable=too-many-locals, too-many-branches, too-ma
...
-            if "certification" in model_kwargs
-            else certification or {"details": "This table is produced by dbt"}
-        )
+        certification_info = {
+            "certification": (
+                model_kwargs.get("extra", {}).pop("certification")
+                if "certification" in model_kwargs.get("extra", {})
+                else certification or {"details": "This table is produced by dbt"}
+            ),
+        }
```
betodealmeida, line 95: Just a nit, but sometimes it's cleaner to use a `try/except`:

```python
try:
    certification_info = model_kwargs["extra"]["certification"]
except KeyError:
    certification_info = certification or DEFAULT_CERTIFICATION
```

(and we'd define `DEFAULT_CERTIFICATION`)

betodealmeida (approved):

src/preset_cli/cli/superset/sync/dbt/datasets.py
Vitor-Avila: That's a very good feedback, @betodealmeida! I'm going to merge this PR for now to solve the issue in `main`, and address this in a future PR.

Vitor-Avila: Handled this here: https://github.com/preset-io/backend-sdk/pull/221

Issue #208, opened 2023-05-11
dbt exposure broken references


advincze-hivemq:
when syncing exposures to dbt, charts that are not built on synced datasets are added with 
```
 depends_on:
      - ref(my_model)
```
instead of 
```
depends_on:
      - ref('my_model')
```

Vitor-Avila:
Ohhh, got it to reproduce this one as well. Going to work on it too.

Vitor-Avila:
I have a potential fix here: https://github.com/preset-io/backend-sdk/pull/221

Issue #207, opened 2023-05-11
synced dbt exposures name need to be an identifier


advincze-hivemq:
rev: `8c7a4b81bd4e1f94aaeb504b278fe7a12ff7e820`

currently the synced exposures from  are added in the following way:
```
version: 2
exposures:
- name: <chart name>
  type: analysis
  maturity: low
  url: <chart url>
  description: ''
  depends_on:
  - ref('...')
  owner:
    name: Bob
    email: unknown
...
```

however the exposure name needs to be an identifier (see https://docs.getdbt.com/docs/build/exposures#declaring-an-exposure)
in snake case so that it can be also used from the dbt docs lineage

the id could be used as a `name` instead and the chart name can be added as the `label` field

dweaver33:
We are also experiencing issues with exposures.  Same as mentioned here and #206

Vitor-Avila:
I have a potential fix here: https://github.com/preset-io/backend-sdk/pull/221

Vitor-Avila:
I just merged https://github.com/preset-io/backend-sdk/pull/221, so going to close this issue. @advincze-hivemq let me know if you still face issues after upgrading the CLI based on the repository. I'm going to work on releasing a new version in PyPI next week.

Issue #206, opened 2023-05-11
syncing dbt exposures fail 


advincze-hivemq:
using rev `8c7a4b81bd4e1f94aaeb504b278fe7a12ff7e820`

sometimes charts have an outdated "query_context" 

The error goes away if you open the chart and it "updates" itself

this is the error I see:
```bash
preset-cli --workspaces <my workspace>  superset sync dbt target/manifest.json --project=<my_project> --target=prod --exposures-only --exposures models/preset_exposures.yml

Traceback (most recent call last):
  File "/Users/adam/dwh/.venv/bin/preset-cli", line 8, in <module>
    sys.exit(preset_cli())
  File "/Users/adam/dwh/.venv/lib/python3.9/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/Users/adam/dwh/.venv/lib/python3.9/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/Users/adam/dwh/.venv/lib/python3.9/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/adam/dwh/.venv/lib/python3.9/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/adam/dwh/.venv/lib/python3.9/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/adam/dwh/.venv/lib/python3.9/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/adam/dwh/.venv/lib/python3.9/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/Users/adam/dwh/.venv/lib/python3.9/site-packages/click/decorators.py", line 26, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/Users/adam/dwh/.venv/lib/python3.9/site-packages/preset_cli/cli/superset/main.py", line 118, in new_command
    ctx.invoke(command, *args, **kwargs)
  File "/Users/adam/dwh/.venv/lib/python3.9/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
[...]

PR #221, opened 2023-06-20
(fix)Exposures definitions and added ability to preserve columns

Vitor-Avila:
This PR addresses a few issues with the exposures sync:
* In the past, we were not including the single quotes to specify the model name [here](https://github.com/preset-io/backend-sdk/blob/main/src/preset_cli/cli/superset/sync/dbt/command.py#L137), which could cause https://github.com/preset-io/backend-sdk/issues/208
* The asset titles were used as the `name` for exposures, which could contain [incompatible characters](https://docs.getdbt.com/reference/exposure-properties): https://github.com/preset-io/backend-sdk/issues/207 and https://github.com/preset-io/backend-sdk/issues/177
* Charts don't have a `query_context` value until they are viewed in Explore, which could cause: https://github.com/preset-io/backend-sdk/issues/206

This PR also introduces a new flag to the dbt sync: ``--preserve-columns``. By default, the sync would update existing datasets in Superset with ``override_columns=True``, which resets the column definition (`groupby` and `filterable` configurations) and also sync columns from the dataset. This could potentially misconfigure existing columns in Superset, or bring back to the datasets columns that were intentionally excluded. This behavior will be kept by default (since the main use-case for the dbt integration is to have the dbt info as the source of truth), however if the ``--preserve-columns`` flag is included in the dbt sync command, the update dataset API calls would include ``override_columns=False``, preserving column configurations and avoiding a column sync.

### DIFF
--- a/src/preset_cli/cli/superset/sync/dbt/command.py
+++ b/src/preset_cli/cli/superset/sync/dbt/command.py
@@ -67,6 +67,12 @@ from preset_cli.exceptions import DatabaseNotFoundError
     default=False,
     help="Do not sync models to datasets and only fetch exposures instead",
 )
+@click.option(
+    "--preserve-columns",
+    is_flag=True,
+    default=False,
+    help="Preserve column configurations",
+)
 @click.pass_context
 def dbt_core(  # pylint: disable=too-many-arguments, too-many-locals
     ctx: click.core.Context,
@@ -81,6 +87,7 @@ def dbt_core(  # pylint: disable=too-many-arguments, too-many-locals
     disallow_edits: bool = True,
     external_url_prefix: str = "",
     exposures_only: bool = False,
+    preserve_columns: bool = False,
 ) -> None:
     """
     Sync models/metrics from dbt Core to Superset and charts/dashboards to dbt exposures.
@@ -88,6 +95,8 @@ def dbt_core(  # pylint: disable=too-many-arguments, too-many-locals
     auth = ctx.obj["AUTH"]
     url = URL(ctx.obj["INSTANCE"])
     client = SupersetClient(url, auth)
+
+    reload_columns = not preserve_columns
 
     if profiles is None:
         profiles = os.path.expanduser("~/.dbt/profiles.yml")
@@ -134,7 +143,7 @@ def dbt_core(  # pylint: disable=too-many-arguments, too-many-locals
             models.append(model_schema.load(config))
     models = apply_select(models, select, exclude)
     model_map = {
-        ModelKey(model["schema"], model["name"]): f'ref({model["name"]})'
+        ModelKey(model["schema"], model["name"]): f'ref(\'{model["name"]}\')'
         for model in models
     }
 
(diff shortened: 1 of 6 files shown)


src/preset_cli/cli/superset/sync/dbt/command.py
```diff
@@ -134,7 +143,7 @@ def dbt_core(  # pylint: disable=too-many-arguments, too-many-locals
             models.append(model_schema.load(config))
     models = apply_select(models, select, exclude)
     model_map = {
-        ModelKey(model["schema"], model["name"]): f'ref({model["name"]})'
+        ModelKey(model["schema"], model["name"]): f'ref(\'{model["name"]}\')'
```
betodealmeida, line 146: Nit:

```suggestion
        ModelKey(model["schema"], model["name"]): f"ref('{model['name']}')"
```

src/preset_cli/cli/superset/sync/dbt/command.py
```diff
@@ -340,7 +359,7 @@ def dbt_cloud(  # pylint: disable=too-many-arguments, too-many-locals
     models = dbt_client.get_models(job_id)
     models = apply_select(models, select, exclude)
     model_map = {
-        ModelKey(model["schema"], model["name"]): f'ref({model["name"]})'
+        ModelKey(model["schema"], model["name"]): f'ref(\'{model["name"]}\')'
```
betodealmeida, line 362:
```suggestion
        ModelKey(model["schema"], model["name"]): f"ref('{model['name']}')"
```

src/preset_cli/cli/superset/sync/dbt/exposures.py
```diff
@@ -110,8 +120,21 @@ def sync_exposures(  # pylint: disable=too-many-locals
...
+        asset_title = re.sub(
+            r"[^a-zA-Z0-9 _]",
+            "",
+            chart["slice_name"],
+        )  # remove unsupported characters
+        asset_title = re.sub(
+            r" ",
+            "_",
+            asset_title,
+        )  # replace blank spaces with underscores
```
betodealmeida, line 133: `\W` is equivalent to `[^a-zA-Z0-9_]` (no space, though), so if you replace spaces with underscores first you can simplify this:

```suggestion
        asset_title = re.sub(" ", "_", chart["slice_name"])
        asset_title = re.sub("\W", "", asset_title)
```

src/preset_cli/cli/superset/sync/dbt/exposures.py
```diff
@@ -131,8 +154,21 @@ def sync_exposures(  # pylint: disable=too-many-locals
...
+        asset_title = re.sub(
+            r"[^a-zA-Z0-9 _]",
+            "",
+            dashboard["dashboard_title"],
+        )  # remove unsupported characters
+        asset_title = re.sub(
+            r" ",
+            "_",
+            asset_title,
+        )  # replace blank spaces with underscores
```
betodealmeida, line 167: Same here:

```suggestion
        asset_title = re.sub(" ", "_", dashboard["dashboard_title"])
        asset_title = re.sub("\W", "", asset_title)
```

betodealmeida (approved):

src/preset_cli/cli/superset/sync/dbt/exposures.py
Vitor-Avila: ohhhhh, didn't know that! Thanks for sharing!

PR #222, opened 2023-06-22
(fix)get_role_id() was returning multiple results

Vitor-Avila:
The `get_role_id()` function was previously configured to use ``{"_flt_0_name": role_name}``. The ``0`` filter index is used for **Starts with**, so in case there are more than one DARs starting with the same name from the YAML file, multiple results would be returned.

This PR changes the filter index to ``3`` which is used for the **Equals** filter operator. The ``3`` filter operator was already being used [here](https://github.com/preset-io/backend-sdk/blob/main/src/preset_cli/api/clients/superset.py#L920), so this should be a minor change.


betodealmeida (approved):
Thanks!!!

PR #225, opened 2023-07-14
fix(dbt-cloud-sync): Make the Schemas more flexible

Vitor-Avila:
Adjustments to dbt marshmallow schemas to avoid integration errors. I noticed that it's possible some repository-related infos will be returned as `null`, and also that the `docs_job` and `freshness_job` actually return a `Job` object. Did a few modifications to the `Job` object for compatibility in both cases.


betodealmeida (commented):
Nice!

betodealmeida (approved):

Issue #223, opened 2023-07-03
The Export RLS command is not working properly 


Vitor-Avila:
The Export RLS command is generating an empty YAML file. I believe this could be related with the React/REST revamp of this functionality: https://github.com/apache/superset/pull/22325

Vitor-Avila:
I have a tentative fix here: https://github.com/preset-io/backend-sdk/pull/227

PR #227, opened 2023-07-18
fix(export-rls): Support Preset and older Superset versions

Vitor-Avila:
This PR improves the `export-roles` command, so that it supports Preset (and most recent Superset versions) since this was migrated to react and APIs. It also support older Superset instances in which RLS don't have a `name` and `description`.

This addresses https://github.com/preset-io/backend-sdk/issues/223.


betodealmeida (approved):
Beautiful!

PR #229, opened 2023-07-26
Making dbt schemas more flexible

Vitor-Avila:
Marking more fields as nullable based on the dbt API docs https://docs.getdbt.com/dbt-cloud/api-v2


betodealmeida (approved):
Thanks!

Issue #231, opened 2023-07-31
dbt sync is failing to create virtual datasets for Preset Cloud


Vitor-Avila:
### Description
When triggering a dbt sync from database engines that support having multiple DBs (like Snowflake), models powered from tables that aren't from the DB connected to Preset are created as virtual. This process is currently not working because the CLI is using deprecated endpoints (https://github.com/preset-io/backend-sdk/issues/219).

### How to reproduce the bug
1. Connect dbt to a DB engine that supports having multiple DBs (such as Snowflake). 
2. Connect DB A to Preset.
3. Create a model powered by DB B in dbt.
4. Sync this model to Preset.

### Expected behavior
A virtual dataset should be created.

### Actual behavior
The operation fails

### Workaround
No workarounds are currently known.

Issue #219, opened 2023-06-16
Migrate `/superset/sql_json/` to `/api/v1/sqllab/execute`


cwegener:
Probably not a feature that many people use, but now that `/superset/sql_json/` is gone, the SQL execution CLI will stop working until migrated to the new sqllab API.

https://github.com/preset-io/backend-sdk/blob/c95fd10965c8585407101f62db8c54cf07915a62/src/preset_cli/api/clients/superset.py#L243-L277

PR #232, opened 2023-07-31
fix(dbt-sync): Migrate legacy endpoints to fix virtual dataset creation

Vitor-Avila:
This PR updates the `_run_query` function to use the new API endpoint by default, but maintains backwards compatibility in case the new endpoint is not available (returns 404) (for older Superset versions).

By default, the virtual dataset creation would include running the query, but this PR also updates the default behavior since now the dataset endpoint supports virtual dataset creation directly (without `columns` information). This change was also made with backwards compatibility in mind for older Superset versions.

This PR fixes https://github.com/preset-io/backend-sdk/issues/231 and https://github.com/preset-io/backend-sdk/issues/219.

### DIFF
--- a/src/preset_cli/api/clients/superset.py
+++ b/src/preset_cli/api/clients/superset.py
@@ -269,17 +269,16 @@ class SupersetClient:  # pylint: disable=too-many-public-methods
         }
         self.session.headers.update(headers)
 
-        try:
+        _logger.debug("POST %s\n%s", url, json.dumps(data, indent=4))
+        response = self.session.post(url, json=data)
+        
+        # Legacy superset installations don't have the SQL API endpoint yet
+        if response.status_code == 404:
+            url = self.baseurl / "superset/sql_json/"
             _logger.debug("POST %s\n%s", url, json.dumps(data, indent=4))
             response = self.session.post(url, json=data)
-            response.raise_for_status()
-
-        # Legacy superset installations don't have the SQL API endpoint yet
-        except requests.exceptions.HTTPError as err:
-            if err.response.status_code == 404:
-                url = self.baseurl / "superset/sql_json/"
-                _logger.debug("POST %s\n%s", url, json.dumps(data, indent=4))
-                response = self.session.post(url, json=data)
+            
+        response.raise_for_status()
 
         validate_response(response)
         payload = response.json()
@@ -590,6 +589,7 @@ class SupersetClient:  # pylint: disable=too-many-public-methods
         for column in columns:
             column["column_name"] = column["name"]
             column["groupby"] = True
+            # Superset 1.4 returns ``is_date`` instead of ``is_dttm``
             if column.get("is_dttm") or column.get("is_date"):
                 column["type_generic"] = 2
             elif column["type"] is None:
@@ -615,6 +615,7 @@ class SupersetClient:  # pylint: disable=too-many-public-methods
 
         payload = response.json()
 
+        # Superset 1.4 returns ``{"table_id": dataset_id}`` rather than the dataset payload
         return payload["data"] if "data" in payload else {"id": payload["table_id"]}
 
     def update_dataset(
--- a/src/preset_cli/api/clients/superset.py
+++ b/src/preset_cli/api/clients/superset.py
@@ -43,6 +43,7 @@ from zipfile import ZipFile
 
 import pandas as pd
 import prison
+import requests
 import yaml
 from bs4 import BeautifulSoup
 from yarl import URL
@@ -247,7 +248,7 @@ class SupersetClient:  # pylint: disable=too-many-public-methods
         schema: Optional[str] = None,
         limit: int = 1000,
     ) -> Dict[str, Any]:
-        url = self.baseurl / "superset/sql_json/"
+        url = self.baseurl / "api/v1/sqllab/execute/"
         data = {
             "client_id": shortid()[:10],
             "database_id": database_id,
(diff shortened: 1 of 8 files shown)


src/preset_cli/api/clients/superset.py
```diff
@@ -526,6 +567,16 @@ def create_dataset(self, **kwargs: Any) -> Any:
...
 
+        # Check if the dataset creation supports sql directly
+        not_legacy = self.validate_key_in_resource_schema(
+            "dataset",
+            "sql",
+            keys=["add_columns"],
+        )
+        not_legacy = not_legacy["add_columns"]
+        if not_legacy:
+            return self.create_resource("dataset", **kwargs)
```
Vitor-Avila, line 578: I thought about doing the validation first, and then change line 567 to `if "sql" not in kwargs or not_legacy` so that we would only have a single `return`, but then I realized most commonly the integration syncs physical datasets, and this change would always execute `validate_key_in_resource_schema` which is another API request. With that in mind, I decided to have a second block for this, so that we only call this other endpoint if needed (trying to not increase the dataset sync duration).

src/preset_cli/api/clients/superset.py
```diff
@@ -539,7 +590,7 @@ def create_dataset(self, **kwargs: Any) -> Any:
         for column in columns:
             column["column_name"] = column["name"]
             column["groupby"] = True
-            if column["is_dttm"]:
+            if column.get("is_dttm") or column.get("is_date"):
```
Vitor-Avila, line 593: Did some tests using Superset `1.4` and noticed that it returns `is_date` rather than `is_dttm`.

src/preset_cli/api/clients/superset.py
```diff
@@ -564,7 +615,7 @@ def create_dataset(self, **kwargs: Any) -> Any:
 
         payload = response.json()
 
-        return payload["data"]
+        return payload["data"] if "data" in payload else {"id": payload["table_id"]}
```
Vitor-Avila, line 618: Similarly, Superset `1.4` returns `{"table_id": dataset_id}` rather than the dataset payload. We only use the `id` for now, so not returning more info for legacy instances shouldn't be a blocker.

src/preset_cli/cli/superset/sync/dbt/command.py
```diff
@@ -84,7 +84,7 @@ def dbt_core(  # pylint: disable=too-many-arguments, too-many-locals
     profiles: Optional[str] = None,
     exposures: Optional[str] = None,
     import_db: bool = False,
-    disallow_edits: bool = True,
+    disallow_edits: bool = False,
```
Vitor-Avila, line 87: I noticed this was set to `True` here even tough the default state is `False`, so I updated only for consistency but it doesn't really change the result. It should be totally fine to revert it to `True` if we want to.

src/preset_cli/cli/superset/sync/dbt/command.py
```diff
@@ -102,6 +102,17 @@ def dbt_core(  # pylint: disable=too-many-arguments, too-many-locals
...
+
+    if "MANAGER_URL" not in ctx.obj and disallow_edits:
+        warnings.warn(
+            (
+                "The managed externally feature was only introduced in Superset v1.5."
+                "Make sure you are running a compatible version."
+            ),
+            category=UserWarning,
+            stacklevel=2,
+        )
```
Vitor-Avila, line 114: Thought about only including `is_managed_externally` to the update payloads in case `--disallow-edits` flag was used, but then I realized that if a user wanted to trigger a new sync without this flag to set it to `False`, that wouldn't work. So I only added this warning which in addition to the error faced for legacy instances should be pretty straightforward (the error would actually happens even if `--disallow-edits` isn't included).

src/preset_cli/api/clients/superset.py
```diff
@@ -268,10 +269,19 @@ def _run_query(
...
+            _logger.debug("POST %s\n%s", url, json.dumps(data, indent=4))
+            response = self.session.post(url, json=data)
+            response.raise_for_status()
 
+        # Legacy superset installations don't have the SQL API endpoint yet
+        except requests.exceptions.HTTPError as err:
+            if err.response.status_code == 404:
+                url = self.baseurl / "superset/sql_json/"
+                _logger.debug("POST %s\n%s", url, json.dumps(data, indent=4))
+                response = self.session.post(url, json=data)
```
betodealmeida, line 282: What happens if we have an error but it's not 404? We need to raise the exception if something other than a 404 happens. This can be simplified a little bit to do that:

```suggestion
        _logger.debug("POST %s\n%s", url, json.dumps(data, indent=4))
        response = self.session.post(url, json=data)
        
        # Legacy superset installations don't have the SQL API endpoint yet
        if response.status_code == 404:
            url = self.baseurl / "superset/sql_json/"
            _logger.debug("POST %s\n%s", url, json.dumps(data, indent=4))
            response = self.session.post(url, json=data)
            
        response.raise_for_status()
```

src/preset_cli/api/clients/superset.py
betodealmeida:
```suggestion
            # Superset 1.4 returns ``is_date`` instead of ``is_dttm``
            if column.get("is_dttm") or column.get("is_date"):
```

src/preset_cli/api/clients/superset.py
betodealmeida, line 618:
```suggestion
        # Superset 1.4 returns ``{"table_id": dataset_id}`` rather than the dataset payload
        return payload["data"] if "data" in payload else {"id": payload["table_id"]}
```

betodealmeida (approved):
This is amazing! I left a few suggestions.

src/preset_cli/api/clients/superset.py
Vitor-Avila: that's totally right! Thanks for catching that!

Vitor-Avila: I don't think we need the `raise_for_status()` anymore so I've removed it as well.

PR #233, opened 2023-08-07
fix(superset-auth): Retrieve CSRF token with jwt auth for Superset

### LINKED ISSUE #183: Calling superset assets api with JWT token fails because no csrf token is used
Trying to use backend sdk with a superset instance using the jwt authentication. The action is import. 
Problem is, call to assets api fails since it's a POST call but it's not sending a csrf token. 
Could be useful to fix that.
Thanks,
Roy

Vitor-Avila:
By default, Superset requires an `X-CSRFToken` header for some API requests. This was already handled when using the `superset-cli` command with basic auth (`-u $username -p $password`) but it wasn't handled when using jwt auth (`--jwt-token $jwt`).

This PR adds logic to include the `CSRF` token in the `get_headers` function. 

Fixes https://github.com/preset-io/backend-sdk/issues/183.


tests/cli/superset/main_test.py
```diff
@@ -161,4 +162,4 @@ def test_superset_jwt_auth(mocker: MockerFixture) -> None:
         catch_exceptions=False,
     )
 
-    JWTAuth.assert_called_with("SECRET")
+    JWTAuth.assert_called_with("SECRET", URL("http://localhost:8088/"))
```
zephyring, line 165: baseurl is better suffix without `/`. Then you can prefix `api/v1/security/csrf_token/` with `/`

Vitor-Avila: thanks, @zephyring! The `baseurl` wasn't included in the constructor before, so I just copied what was sent to the `runner.invoke` [here](https://github.com/preset-io/backend-sdk/blob/main/tests/cli/superset/main_test.py#L158):
``` python
runner.invoke(
  superset_cli,
  ["--jwt-token=SECRET", "http://localhost:8088/", "export"],
  catch_exceptions=False,
)
```

The last trailing slash is included in other tests (like [here](https://github.com/preset-io/backend-sdk/blob/main/tests/auth/password_test.py#L23) and [here](https://github.com/preset-io/backend-sdk/blob/main/tests/auth/password_test.py#L48)), do you think this should be updated in other places as well? 

I performed a local test using both `http://localhost:8088/` and `http://localhost:8088` in the CLI command and printed the endpoint -- both scenarios produced the same URL (`http://localhost:8088/api/v1/security/csrf_token/`) and the command worked properly. 

Let me know your thoughts. Thank you!

zephyring: It's a convention to have base_url ending without `/`. Not sure the other examples syntax for string / string is a good pattern to follow. maybe @betodealmeida ?

betodealmeida: It should work either way here, we're using [yarl](https://pypi.org/project/yarl/) for URL composition and it should do the right thing.

src/preset_cli/auth/token.py
```diff
@@ -12,9 +14,27 @@ class TokenAuth(Auth):  # pylint: disable=too-few-public-methods, abstract-metho
     Auth via a token.
     """
 
-    def __init__(self, token: str):
+    def __init__(self, token: str, baseurl: Optional[URL] = None):
         super().__init__()
         self.token = token
+        self.baseurl = baseurl
+
+    def get_csrf_token(self, jwt: str) -> str:
```
betodealmeida, line 22: It's better to create a new `Auth` class, maybe `SupersetJWTAuth`, since the path `"api/v1/security/csrf_token/"` is very specific to Superset. Otherwise this generic class becomes tightly coupled with Superset and it becomes hard to reuse it for other APIs.

betodealmeida (changes requested):

src/preset_cli/auth/token.py
Vitor-Avila: ohhhhh that makes sense! Just noticed this class is also used by dbt. Going to make these changes. Thank you!

betodealmeida (approved):

PR #235, opened 2023-08-21
fix(session): Implement retry mechanism for closed connections

Vitor-Avila:
By default, we create a session with the target server and re-use it during the process. However, the server might close the connection, so this PR introduces a `Retry` mechanism to prevent errors.


betodealmeida (approved):
Let's give it a try!

Issue #217, opened 2023-06-07
Allow dbt model selection based on provided path


wojciech-mazur-mlg:
We would like to see dbt model selector based on provided path to run all models located in this path.

Example command:

`preset-cli superset sync dbt-core target/manifest.json --select models/subdirectory`

- It should work the same as for dbt, e.g.: `dbt compile --select models/subdirectory`
- `models/subdirectory` and `models/subdirectory/*` should be equivalent.

Currently it raises NotImplementedError and redirects to ticket creation.

It is somehow similar to issue: https://github.com/preset-io/backend-sdk/issues/171 but this tickets extends that one.

Vitor-Avila:
@wojciech-mazur-mlg I just merged this PR that should implement this requirement: https://github.com/preset-io/backend-sdk/pull/242

As a heads up, using `models/subdirectory/*` directly wouldn't work because the terminal automatically unpacks the folder and adds all files as elements to the CLI command, which would fail since they don't have the `--select` flag to work properly. You can either use quotes (`"models/subdirectory/*"`) to implement this behavior, use `models/subdirectory` or `models/subdirectory/`.

Thanks!

wojciech-mazur-mlg:
Thank you @Vitor-Avila !

Issue #171, opened 2022-12-15
Allow selection of dbt models based on file name


dweaver33:
Currently the `preset-cli` allows selection of dbt models using the graph selectors and tags.  However, it would be nice to also allow the use of the model file names as well.  

Example would be:
```bash
preset-cli ... --select models/datamarts/my_model.sql                                                  #sync only "my_model"
preset-cli ... --select models/datamarts/my_model.sql  --select models/datamarts/my_other_model.sql    #sync "my_model" and "my_other_model"
```

dweaver33:
@Vitor-Avila, any update if we can get this?  Without this, our CI/CD deployments are becoming very time consuming as we have to deploy all the models.

Vitor-Avila:
hey @dweaver33, I've performed a few tests on my end, and I can use `--select` with the actual model names. Are you looking to use the file name instead to use the terminal auto-complete functionality? Also, are you looking to use a folder to include subfiles (https://github.com/preset-io/backend-sdk/issues/217)?

dweaver33:
> hey @dweaver33, I've performed a few tests on my end, and I can use `--select` with the actual model names. Are you looking to use the file name instead to use the terminal auto-complete functionality? Also, are you looking to use a folder to include subfiles (https://github.com/preset-io/backend-sdk/issues/217)?

Yes, I'd like to use the full file name and path as a selection option. A folder option would be nice as well, but it a specific need I have. 

My use case is that we have a process that detects which files have changed. This process outputs the full file names of all files that changes (with a separator). I'd like to feed this output to the Preset CLI so I could selectively deploy models to Preset.

Vitor-Avila:
ok, thanks for sharing more context -- this is really helpful. I'm finishing working on two other CLI issues, and then I can take a look at this.

Vitor-Avila:
@dweaver33 I have a potential implementation for this here: https://github.com/preset-io/backend-sdk/pull/242

PR #242, opened 2023-09-08
chore(dbt-sync): Allow the specification of dbt models using their path

Vitor-Avila:
This PR adds the ability to specify the models that should be synced using their relative path. Both specifying files (for example `--select models/my_model.sql`) and folders (for example `--select models/prod`) work. When selecting a folder, it supports the inclusion or not of the last trailing slash(`--select models/prod == --select models/prod/`). 

It's also possible to use `*` after the folder name, however for this approach it's required to pass the argument as a string -> `--select "models/prod/*"` otherwise the terminal automatically creates one argument per file, breaking the CLI.

Implements https://github.com/preset-io/backend-sdk/issues/217 and https://github.com/preset-io/backend-sdk/issues/171.

### DIFF
--- a/src/preset_cli/cli/superset/sync/dbt/lib.py
+++ b/src/preset_cli/cli/superset/sync/dbt/lib.py
@@ -284,6 +284,7 @@ def load_profiles(
     return apply_templating(profiles)
 
 
+# pylint: disable=R0911
 def filter_models(models: List[ModelSchema], condition: str) -> List[ModelSchema]:
     """
     Filter a list of dbt models given a select condition.
--- a/src/preset_cli/cli/superset/sync/dbt/lib.py
+++ b/src/preset_cli/cli/superset/sync/dbt/lib.py
@@ -310,6 +310,18 @@ def filter_models(models: List[ModelSchema], condition: str) -> List[ModelSchema
     if condition in model_names:
         return [model_names[condition]]
 
+    # file
+    file_path = Path(condition)
+    if file_path.is_file() and file_path.stem in model_names:
+        return [model_names[condition]]
+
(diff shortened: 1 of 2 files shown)


dweaver33:
Would this allow us to pass the file list like this?

`--select models/my_model.sql models/my_model2.sql models/my_model3.sql`

Vitor-Avila:
> Would this allow us to pass the file list like this?
> 
> `--select models/my_model.sql models/my_model2.sql models/my_model3.sql`

@dweaver33 not really, you would have to add one `--select` tag per file, or combine the files in a directory and use the directory instead:

``` bash
--select models/my_model.sql --select models/my_model2.sql --select models/my_model3.sql
```

dweaver33:
> > Would this allow us to pass the file list like this?
> > `--select models/my_model.sql models/my_model2.sql models/my_model3.sql`
> 
> @dweaver33 not really, you would have to add one `--select` tag per file, or combine the files in a directory and use the directory instead:
> 
> ```shell
> --select models/my_model.sql --select models/my_model2.sql --select models/my_model3.sql
> ```

Ah, ok.  I think I can work with this limitation.

betodealmeida (approved):
This is awesome!

dweaver33:
@Vitor-Avila, what's the timeline for cutting a release with this PR in it?  We are eagerly awaiting.  😄

Vitor-Avila:
@dweaver33 wait no more! Just released version `0.2.8` including this change. Let me know if you face any issues to use this functionality.

dweaver33:
Thank you!  We'll be testing over the next few weeks

PR #255, opened 2024-01-05
chore: use sqlglot

betodealmeida:
Switch from `sqlparse` to `sqlglot`. I'm doing this because the support for dbt 1.6 will be based on `sqlglot`.

Unit tests pass.


craig-rueda (approved):
LGTM

PR #256, opened 2024-01-08
feat(dbt): sync metrics from MetricFlow

betodealmeida:
This PR introduces an algorithm for syncing dbt metrics defined using [MetricFlow](https://docs.getdbt.com/docs/build/sl-getting-started). In order to sync metrics it:

1. Queries a new GraphQL endpoint to get all the names of the metrics
2. Queries a new GraphQL endpoint to get the SQL for each of the metrics
3. Converts the SQL (a full query) into a projection, using an algorithm similar to the one used with older metrics

Because in MetricFlow metrics are a higher concept, not all metrics can be converted. For example, a metric could be "sales in France", having an implicit `JOIN` in them. Metrics with joins and/or multiple tables are ignored, and the error is logged.

Note that MetricFlow has a concept called "measure" which is much closer to the Superset metric. Unfortunately the GraphQL API doesn't expose all the information we need about measures to convert them into Superset metrics. Namely, the models to which each measure belong are not available via the API.

Testing:

```bash
$ preset-cli --workspaces=https://b5eda41a.us1a.app.preset.io/ superset sync dbt-cloud dbtc_XXX 485502
```

Models and metrics are synced correctly to: https://b5eda41a.us1a.app.preset.io/tablemodelview/list/?pageIndex=0&sortColumn=changed_on_delta_humanized&sortOrder=desc

### DIFF
--- a/src/preset_cli/api/clients/dbt.py
+++ b/src/preset_cli/api/clients/dbt.py
@@ -611,7 +611,7 @@ class MFMetricSchema(PostelSchema):
     type = PostelEnumField(MFMetricType)
 
 
-class MFSQLEngineSchema(str, Enum):
+class MFSQLEngine(str, Enum):
     """
     Databases supported by MetricFlow.
     """
@@ -630,7 +630,7 @@ class MFMetricWithSQLSchema(MFMetricSchema):
     """
 
     sql = fields.String()
-    dialect = PostelEnumField(MFSQLEngineSchema)
+    dialect = PostelEnumField(MFSQLEngine)
     model = fields.String()
 
 
@@ -851,7 +851,7 @@ class DBTClient:  # pylint: disable=too-few-public-methods
 
         return payload["data"]["compileSql"]["sql"]
 
-    def get_sl_dialect(self, environment_id: int) -> str:
+    def get_sl_dialect(self, environment_id: int) -> MFSQLEngine:
         """
         Get the dialect used in the MetricFlow project.
         """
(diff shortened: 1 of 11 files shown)


src/preset_cli/api/clients/dbt.py
```diff
@@ -683,37 +723,46 @@ def get_models(self, job_id: int) -> List[ModelSchema]:
         Fetch all available models.
         """
         query = """
-            query ($jobId: Int!) {
-                models(jobId: $jobId) {
```
betodealmeida, line 687: This is being deprecated, so I converted to the recommended query.

src/preset_cli/cli/superset/sync/dbt/metrics.py
```diff
@@ -190,9 +212,145 @@ def get_superset_metrics_per_model(
...
+    aliases = {}
+    for expression in parsed_query.find_all(Alias):
+        alias_name = expression.alias
+        expression_text = expression.this.sql()
+        aliases[alias_name] = expression_text
+
+    return aliases
+
+
+def convert_query_to_projection(sql: str, dialect: MFSQLEngine) -> str:
```
betodealmeida, line 247: This is the crux of the PR, converting the SQL from MetricFlow into a projection. I've added a good number of unit tests covering all the scenarios I saw.

craig-rueda (approved):
LGTM

Vitor-Avila:
hey @betodealmeida, thank you so much for working on this feature and also adding such extensive test coverage! Are these changes compatible with older versions of dbt Cloud (like 1.4 and 1.5)?

PR #258, opened 2024-01-10
chore(dbt): move logic to a function

betodealmeida:
Consolidate some logic in a function, since it will have to be called by manager when syncing dbt metrics.


Vitor-Avila (commented):
LGTM

Vitor-Avila (approved):
LGTM

PR #264, opened 2024-02-28
fix(dbt Cloud): allow passing job ID

betodealmeida:
In https://github.com/preset-io/backend-sdk/commit/b09de649 we changed the way the dbt Cloud job is retrieved, since we need more information than just the ID. In order to get the job information we need the account and the project ID, so it became impossible to run the CLI to sync dbt Cloud without being prompted — there's currently no way to pass an account or a project ID to the command.

With this PR we can now do:

```bash
$ preset-cli \
> --workspaces=https://example.app.preset.io/ \
> superset sync dbt-cloud \
> ${DBT_TOKEN} ${ACCOUNT_ID} ${PROJECT_ID} ${JOB_ID}
```

In order to run the sync under CI/CD.

Additionally, when running the script interactively, it will print the IDs:

```bash
$ preset-cli superset sync dbt-cloud ${DBT_TOKEN}
Choose one or more workspaces (eg: 1-3,5,8-):

# Preset Internal #
✅ (1) The Data Lab
> 1
https://example.app.preset.io/

Using account My Account [id=123] since it's the only one

Choose a project:
(1) Analytics [id=45]
(2) MetricFlow test [id=67]
> 2

Choose a job:
(1) Parse [id=12345]
(2) Build [id=67890]
> 2
```


Vitor-Avila (approved):
Awesome!

PR #265, opened 2024-03-05
feat: Metricflow and dbt Core

betodealmeida:
Add support for Metricflow when using dbt Core sync.

Ran:

```bash
$ preset-cli --workspaces=https://49437123.us1a.app.preset.io \
> superset sync dbt-core ~/Projects/github/jaffle-sl-template/target/manifest.json \
> --project=jaffle_shop --target=dev
```

See datasets and metrics in https://49437123.us1a.app.preset.io.

Will update tests.


Vitor-Avila:
This is awesome!!!

Vitor-Avila:
Followed below steps to test these changes:
1. Created a new PSQL DB.
2. Cloned this template repo: https://github.com/dbt-labs/jaffle-sl-template.
3. Add the changes from this PR to avoid errors: https://github.com/dbt-labs/jaffle-sl-template/pull/62.
4. Configured the DB connection to the new PSQL DB.
5. Replaced `DATETIME` in the configs with `TIMESTAMP` (sice I'm using Postgres rather than Snowflake).
6. Followed the steps outlined in the repo README to set up the dbt environment.

src/preset_cli/cli/superset/sync/dbt/command.py
```diff
@@ -338,7 +348,44 @@ def get_job(
...
+    dialect: MFSQLEngine,
+) -> Optional[MFMetricWithSQLSchema]:
+    """
+    Compute a SL metric using the ``mf`` CLI.
+    """
+    mf_metric_schema = MFMetricWithSQLSchema()
+
+    command = ["mf", "query", "--explain", "--metrics", metric["name"]]
+    try:
+        result = subprocess.run(command, capture_output=True, text=True, check=True)
```
Vitor-Avila, line 363: When there are multiple metrics defined, this process might take some time which would be invisible to the user. It might be interesting to add a logger here so the user can follow up as the metrics are getting evaluated. For example:
``` python
logger.info("Using `mf` command to retrieve SQL syntax for metric %s", metric["name"])
```

src/preset_cli/cli/superset/sync/dbt/command.py
```diff
@@ -338,7 +348,44 @@ def get_job(
...
+    except subprocess.CalledProcessError:
+        return None
```
Vitor-Avila, line 365: Now the CLI requires having `metricflow` installed to properly retrieve the SQL syntax for metrics. I wonder if we can identify the error returned in case it's not installed (maybe something like `command not found` and provide an error to the user instructing to install the package. We could use `sys.exit(1)` to make sure a CI job would also show as failed so the user adds the specific `metricflow` package (according to their engine) to their configurations.

We could also add the package to the list of requirements.

src/preset_cli/cli/superset/sync/dbt/command.py
```diff
@@ -338,7 +348,44 @@ def get_job(
...
+
+    output = result.stdout.strip()
+    start = output.find("SELECT")
+    sql = output[start:]
+
+    models = get_models_from_sql(sql, dialect, model_map)
+    if len(models) > 1:
+        return None
```
Vitor-Avila, line 373: I'm working on this PR https://github.com/preset-io/backend-sdk/pull/266 and could potentially further improve it to also list at the end of the execution any failed metric (according to the limitations we currently have).

Vitor-Avila (approved):
Thank you so much for working on this! I left some non-blocking comments.

PR #266, opened 2024-03-05
feat(dbt): Adding flag to raise an error when a model sync fails

Vitor-Avila:
Currently, during the dbt sync models might fail to be synced. The operation is not atomic (a sync failure won't prevent other models from being synced) and the CLI logs the error messages. 

While the error messages are helpful to debug the issue, when using the CLI in a CI/CD pipeline, the fact that the execution ends up successfully would show the job in a successful state, which could be misleading since there were issues. 

This PR adds a new flag `--raise-failures` that when used would end the command execution with an error status code in case any model failed to sync.

### DIFF
--- a/src/preset_cli/cli/superset/sync/dbt/command.py
+++ b/src/preset_cli/cli/superset/sync/dbt/command.py
@@ -198,8 +198,8 @@ def dbt_core(  # pylint: disable=too-many-arguments, too-many-branches, too-many
             models.append(model_schema.load(config))
     models = apply_select(models, select, exclude)
     model_map = {ModelKey(model["schema"], model["name"]): model for model in models}
-    
-    failures = []
+
+    failures: List[str] = []
 
     if exposures_only:
         datasets = [
@@ -247,7 +247,7 @@ def dbt_core(  # pylint: disable=too-many-arguments, too-many-branches, too-many
     if exposures:
         exposures = os.path.expanduser(exposures)
         sync_exposures(client, Path(exposures), datasets, model_map)
-    
+
     if failures and raise_failures:
         list_failed_models(failures)
 
@@ -520,7 +520,7 @@ def dbt_cloud(  # pylint: disable=too-many-arguments, too-many-locals
     sl_metrics = process_sl_metrics(dbt_client, job["environment_id"], model_map)
     superset_metrics = get_superset_metrics_per_model(og_metrics, sl_metrics)
 
-    failures = []
+    failures: List[str] = []
 
     if exposures_only:
         datasets = [
--- a/src/preset_cli/cli/superset/sync/dbt/command.py
+++ b/src/preset_cli/cli/superset/sync/dbt/command.py
@@ -24,7 +24,7 @@ from preset_cli.auth.token import TokenAuth
 from preset_cli.cli.superset.sync.dbt.databases import sync_database
 from preset_cli.cli.superset.sync.dbt.datasets import sync_datasets
 from preset_cli.cli.superset.sync.dbt.exposures import ModelKey, sync_exposures
-from preset_cli.cli.superset.sync.dbt.lib import apply_select
+from preset_cli.cli.superset.sync.dbt.lib import apply_select, list_failed_models
 from preset_cli.cli.superset.sync.dbt.metrics import (
     get_models_from_sql,
     get_superset_metrics_per_model,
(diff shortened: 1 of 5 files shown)


src/preset_cli/cli/superset/sync/dbt/lib.py
```diff
@@ -459,3 +461,20 @@ def apply_select(
                 del selected[id_]
 
     return list(selected.values())
+
+
+def list_failed_models(failed_models: List[str]) -> None:
+    """
+    List models that failed to sync and ends execution with an error code
+    """
+    error_message = "Below model(s) failed to sync:"
```
Vitor-Avila, line 470: I thought about adding a redundant check here like this:
``` python
if not failed_models:
    return
```

But I thought it wasn't really necessary since we're only calling this method in case `failed_models` has any value. Let me know if you think we should also add this check here.

tests/cli/superset/sync/dbt/lib_test.py
```diff
@@ -619,3 +620,35 @@ def test_apply_select_using_path(fs: FakeFilesystem) -> None:
...
+    expected_output = "Below model(s) failed to sync:\n - single_failure\n"
+
+    assert captured.out == expected_output
+    assert excinfo.type == SystemExit
+    assert excinfo.value.code == 1
+
+
+def test_list_failed_models_multiple_models(capsys) -> None:
+    """
+    Test ``list_failed_models()`` with multipled failed models
```
Vitor-Avila, line 642: Fix typo

betodealmeida (changes requested):
This is awesome! My only comment is that we should replace `raise_error` with a proper `raise Exception`, see my comments.

src/preset_cli/cli/main.py
```diff
@@ -139,13 +139,7 @@ def preset_cli(  # pylint: disable=too-many-branches, too-many-locals, too-many-
...
                     api_secret = credentials["api_secret"]
                 except Exception:  # pylint: disable=broad-except
-                    click.echo(
-                        click.style(
-                            "Couldn't read credentials",
-                            fg="bright_red",
-                        ),
-                    )
-                    sys.exit(1)
+                    raise_error(1, "Couldn't read credentials")
```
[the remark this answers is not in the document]
Vitor-Avila: ohh that's a good idea! let me look into that!

src/preset_cli/cli/main.py
```diff
@@ -629,9 +601,9 @@ def sync_user_role_to_workspace(
     )
 
 
-preset_cli.add_command(auth)
-preset_cli.add_command(invite_users)
-preset_cli.add_command(import_users)
+preset_cli.add_command(auth, name="auth")
+preset_cli.add_command(invite_users, name="invite-users")
+preset_cli.add_command(import_users, name="import-users")
```
Vitor-Avila, line 606: For some reason, for the commands that involved the `raise_cli_errors` decorator, the tests were failing with command not found until I explicitly added the command names. 🤷

src/preset_cli/cli/superset/sync/dbt/command.py
```diff
@@ -112,25 +119,24 @@ def dbt_core(  # pylint: disable=too-many-arguments, too-many-branches, too-many
...
 ) -> None:
     """
     Sync models/metrics from dbt Core to Superset and charts/dashboards to dbt exposures.
     """
     auth = ctx.obj["AUTH"]
     url = URL(ctx.obj["INSTANCE"])
     client = SupersetClient(url, auth)
+    deprecation_notice: bool = False
+    if raise_failures:
+        warnings.simplefilter("always", DeprecationWarning)
```
Vitor-Avila, line 132: I added this otherwise the `DeprecationWarning` doesn't get logged in the terminal.

src/preset_cli/lib.py
```diff
@@ -40,6 +42,7 @@ def setup_logging(loglevel: str) -> None:
         handlers=[RichHandler()],
         force=True,
     )
+    logging.captureWarnings(True)
```
Vitor-Avila, line 45: Also add to add this to include `DeprecationWarnings` in the terminal

.pre-commit-config.yaml
```diff
@@ -54,7 +54,7 @@ repos:
   #  additional_dependencies: [flake8-bugbear]
 
 - repo: https://github.com/pre-commit/mirrors-mypy
-  rev: 'v0.910'  # Use the sha / tag you want to point at
+  rev: 'v0.981'  # Use the sha / tag you want to point at
```
Vitor-Avila, line 57: Had to bump it to fix `pre-commit` for newer Python versions. https://github.com/python/mypy/issues/13627

betodealmeida (approved):
Awesome!

PR #267, opened 2024-03-14
fix: Bump YARL dependency to retain trailing slashes in URLs

oliverlambson:


oliverlambson:
@betodealmeida would appreciate a review on this, thanks

Vitor-Avila:
Hey @oliverlambson, thanks for submitting these changes! I'm afraid that due to license restrictions, we're currently not able to accept/merge contributions from outside of Preset. I'll take a look on this scenario and see if we can address this issue. Thanks!

oliverlambson:
> Hey @oliverlambson, thanks for submitting these changes! I'm afraid that due to license restrictions, we're currently not able to accept/merge contributions from outside of Preset. I'll take a look on this scenario and see if we can address this issue. Thanks!

Thanks!

Vitor-Avila:
Hey @oliverlambson, 

Our team has been working to implement this CLA that allows us to accept external contributions to the CLI -- thank you for singing it 🙌 

I was performing some tests around this flow, and I couldn't reproduce the issue in the first place. I'm not opinionated against the proposed fix, but I was wondering if I could reproduce to check if we could add test coverage to prevent any regression in the future. This is how I tried to reproduce it:

1. Created a new Python virtual environment using Python `v3.9.0`.
2. Installed Superset `v3.0.0` through `pip`.
3. Tried using it with the CLI right away (command below). Faced a continuous loop that never really threw any errors (in the CLI). In the Superset logs, I could find `flask_wtf.csrf.CSRFError: 400 Bad Request: The CSRF session token is missing.`.
``` bash
superset-cli --loglevel=DEBUG -u admin -p admin http://localhost:8088 export-assets --chart-ids=101 . --overwrite
```
4. I thought this could be the issue, but then I ran `pip install --upgrade yarl==1.9.4` in the CLI env and I was still facing the same issue.
5. Downgraded `yarl` to the previous version, and then updated below configs in Superset:
``` python
WTF_CSRF_ENABLED = False
SESSION_COOKIE_SAMESITE = None
```
6. Sync command still failing. Upgraded `yarl` again, still same issue.
7. Lastly, set `TALISMAN_ENABLED = False` in the Superset config, and then the sync was back working (with the old `yarl` version).

I also tried using `ngrok` to force an `https` URL but also couldn't reproduce the issue.

Do you have any specific steps I could follow to repro it? Do you know if this could be related with your CSP/Talisman config?

PR #268, opened 2024-03-14
feat(dbt): Supporting sync columns with --preserve-metadata and --merge-metadata

Vitor-Avila:
The `--preserve-metadata` and `--merge-metadata` flags can be used with the dbt sync command so that Superset metadata (specially metrics) is not automatically discarded during the sync. To support this flow, currently these two flags invoke the `update_dataset()` method with `reload_columns=False`. This is intentionally implemented so that the `groupby`, `filterable` and other configurations are not reset to default. 

The issue with this approach is that the dataset columns are not synced, so in case the model was modified to include a new column (or a column was removed) this change won't reflect to Superset automatically. 

This PR updates these two flags so that columns are now synced while still preserving the column metadata. It also refactors the `sync_datasets()` method to break down all required actions into helper methods.


betodealmeida (approved):
Looks great! Thanks for the refactor!

PR #269, opened 2024-03-14
Fix individual dashboard upload when filter datasets are not the same as in charts

grvoicu:
Preset supports dashboards with filters on different datasets than the datasets used by the charts. currently, individually uploading such a dashboard with Preset CLI fails, because the filters datasets are not included in the list of dependencies for the dashboard asset.

Note: I consider having dashboard filters on different datasets than the datasets used by the charts is not a bug, but a feature. An example where this is useful: a dataset with all the countries in the world and multiple datasets, each one with data for a subset of countries. A filter on the world countries dataset can then be applied to any dashboard, without having to take care to know which dataset is used in the dashboard charts. As long as the country column is named the same and the values are consistent across all datasets it will work.

### DIFF
--- a/src/preset_cli/cli/superset/sync/native/command.py
+++ b/src/preset_cli/cli/superset/sync/native/command.py
@@ -11,7 +11,7 @@ from datetime import datetime, timezone
 from io import BytesIO
 from pathlib import Path
 from types import ModuleType
-from typing import Any, Dict, Iterator, Tuple
+from typing import Any, Dict, Iterator, Set, Tuple
 from zipfile import ZipFile
 
 import backoff
@@ -263,7 +263,7 @@ def import_resources_individually(
             ("databases", lambda config: []),
             ("datasets", lambda config: [config["database_uuid"]]),
             ("charts", lambda config: [config["dataset_uuid"]]),
-            ("dashboards", get_charts_uuids),
+            ("dashboards", get_charts_and_filters_uuids),
         ]
         related_configs: Dict[str, Dict[Path, AssetConfig]] = {}
         for resource_name, get_related_uuids in imports:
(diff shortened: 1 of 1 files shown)


Vitor-Avila:
Hi @grvoicu,

Thank you so much for contributing this fix to the CLI. Since the CLI is under an `all-rights-reserved` project, we can only accept external contributions in case users sign a CLA (we have an automated flow through [the PR comments](https://github.com/preset-io/backend-sdk/pull/269#issuecomment-2008169648)).

We also have a 100% test coverage requirement in the repository, so we might need to add a few test cases (and review any test that might break with changes) to be able to merge it. 

Would you like to proceed with this contribution? If so please sign the CLA, and let me know if you would need assistance with the tests. 

Thank you!

Vitor-Avila:
Hey @grvoicu, 

Wondering if you had a chance to check my message. Thank you!

src/preset_cli/cli/superset/sync/native/command.py
```diff
@@ -301,6 +308,19 @@ def get_charts_uuids(config: AssetConfig) -> Iterator[str]:
             yield child["meta"]["uuid"]
 
 
+def get_filters_dataset_uuids(config: AssetConfig) -> Set[str]:
+    """
+    Extract native filters dataset UUID from a dashboard config.
+    """
+    uuids = set()
+    for child in config["metadata"]["native_filter_configuration"]:
```
Mast3rSensei, line 316: This might actually fail for dashboard without either `metadata` nor `native_filter_configuration` attributes. Better to add

`if config.get('metadata') and config['metadata'].get('native_filter_configuration'):`

Mast3rSensei (changes requested):

Vitor-Avila:
Closing in favor of https://github.com/preset-io/backend-sdk/pull/292

PR #270, opened 2024-03-15
fix(dbt): Support syncing derived metrics on legacy dbt versions

Vitor-Avila:
[PR#254](https://github.com/preset-io/backend-sdk/pull/254/files#diff-6a9da321f6c92c5e31f999ebd04d4f2f1c0dceb066d3aae0d59cc836eb2444b4R157-R164) introduced a new pattern for mapping metrics:

``` python
metric_map = {metric["unique_id"]: metric for metric in metrics}
```

This ended up affecting our ability to parse derived metrics, given the metric macro renders the metric `name`, rather than its `unique_id`. As a consequence, [this condition](https://github.com/preset-io/backend-sdk/pull/255/files#diff-6a9da321f6c92c5e31f999ebd04d4f2f1c0dceb066d3aae0d59cc836eb2444b4R81-R82) was never met (since `token.sql()` would return the macro name):

``` python
for token in tokens:
      if token.sql() in metrics:
```

This section of the code is only used by legacy dbt versions -- for new versions we're relying on `MetricFlow` to provide the metric syntax.

This PR re-introduces the old mapping structure specifically for this context:
``` python
metric_map = {metric["name"]: metric for metric in metrics}
```

It also includes the `dialect` as part of the metric schema, to assist `sqlglot` when parsing the syntax. I was facing an issue when trying to parse below metric syntax since it's BQ-specific:
``` sql
SAFE_DIVIDE(
   SUM(
      IF(
         `product_line` = "Classic Cars",
         price_each * 0.80,
         price_each * 0.70
      )
   ),
   revenue_verbose_name_from_dbt
)
```

### DIFF
--- a/src/preset_cli/api/clients/dbt.py
+++ b/src/preset_cli/api/clients/dbt.py
@@ -589,6 +589,7 @@ class MetricSchema(PostelSchema):
     # dbt >= 1.3
     calculation_method = fields.String()
     expression = fields.String()
+    dialect = fields.String()
 
 
 class MFMetricType(str, Enum):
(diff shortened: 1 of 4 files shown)


betodealmeida (approved):
Ah, great catch!

PR #273, opened 2024-03-20
fix(sqlglot): Convert metric syntax back to dialect-specific after parsing it

Vitor-Avila:
The CLI uses `sqlglot` to parse SQL syntax from dbt metrics. While we were specifying the SQL dialect when parsing the query, we were not specifying the dialect when converting the parsed result back to SQL, which would cause issues with engine-specific syntaxes. Some examples:
* A BQ metric created using `IF()` statement would get converted to `CASE` (would still work but different syntax)
* A BQ metric created using `DATE_DIFF()` would get converted to `DATEDIFF()` which doesn't work with BQ.


betodealmeida (approved):
Aaah, good catch!

PR #274, opened 2024-03-21
fix(dbt): Support filtered COUNT DISTINCT in metrics

Vitor-Avila:
When creating a `COUNT (DISTINCT x)` metric with a filter -- for example:
``` yaml
semantic_models:
  - name: datadiff_test
    ...
    measures:
      - name: distinct_count_test
        expr: id
        description: "The count of unique rows."
        agg: count_distinct
    dimensions:
      - name: quantity
        type: categorical
    ...

metrics:
  - name: test_distinct_metric
    description: "The number of unique rows."
    type: simple
    label: Number of CX Sent Secure Messages
    type_params:
      measure: distinct_count_test
    filter: |                                     
      {{  Dimension('id__quantity') }} > 10
```

The metric conversion would result in an incompatible syntax:
``` sql
COUNT(CASE WHEN quantity > 10 THEN DISTINCT id END)
```

This PR updates the logic to replace the `DISTINCT` as needed:
``` sql
COUNT(DISTINCT CASE WHEN quantity > 10 THEN id END)
```

### DIFF
--- a/src/preset_cli/cli/superset/sync/dbt/metrics.py
+++ b/src/preset_cli/cli/superset/sync/dbt/metrics.py
@@ -13,7 +13,17 @@ from typing import Dict, List, Optional, Set
 
 import sqlglot
 from sqlglot import Expression, exp, parse_one
-from sqlglot.expressions import Alias, Case, Identifier, If, Join, Select, Table, Where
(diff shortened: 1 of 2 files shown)


betodealmeida (approved):
Lovely!

PR #277, opened 2024-04-04
fix(dbt-derived-metrics): Support raw metrics and Jinja syntax

Vitor-Avila:
dbt allows the creation of derived metrics with just raw SQL, not involving any other dbt metric:
``` yaml
metrics:
  - name: new_derived_again
    label: test
    calculation_method: derived
    expression: |
      {% raw %}
       (SUM(price_each)) * 1.2
      {% endraw %}
```

These would currently fail to sync, as derived metrics don't have a `model` definition, so currently the CLI identifies the desired model by getting the models used by the metrics used in the syntax (which don't exist in this case).

To support this flow, the CLI now will get the model's `unique_id` from `meta.superset.model` for such metrics:
``` yaml
metrics:
  - name: new_derived_again
    label: test
    calculation_method: derived
    expression: |
      {% raw %}
       (SUM(price_each)) * 1.2
      {% endraw %}
    meta:
      superset:
        model: model.postgres.vehicle_sales
```

the `{% raw %}` macro also allows users to use Superset-specific Jinja in their metrics:
``` yaml
metrics:
  - name: new_derived_again
    label: test
    calculation_method: derived
    expression: |
      {% raw %}
        SUM(
          {% for x in filter_values('x_values') %}
            {{ + x_values }}
          {% endfor %}
        )
      {% endraw %}
    meta:
      superset:
        model: model.postgres.vehicle_sales
  - name: new_derived
    label: test
    calculation_method: derived
    expression: |
      {% raw %}
        SUM(
          {% for x in filter_values('x_values') %}
            {{ + 
      {% endraw %}
              {{ metric('revenue_verbose_name_from_dbt')}}
      {% raw %}
            }}
          {% endfor %}
        )
      {% endraw %}
```

These were currently failing as we rely on `sqlglot` to parse the derived metrics and replace a metric `name` with its sql syntax. This PR changes the logic so that the SQL parsing is skipped in case the derived metric doesn't really include other metrics. In case it's a combination of Superset-Jinja + dbt Jinja (like the second example), the CLI will catch the `ParseError` exception and will perform a replace with regex for the metric (definitely not as stable as SQL parsing, but hoping to increase compatibility with these types of metrics).

These changes are currently aiming older versions of dbt -- going to also perform some tests with `MetricFlow`/new semantic layer to validate if these flows are supported in newer versions, to assess how to also support them with the CLI.

### DIFF
--- a/src/preset_cli/api/clients/dbt.py
+++ b/src/preset_cli/api/clients/dbt.py
@@ -590,6 +590,7 @@ class MetricSchema(PostelSchema):
     calculation_method = fields.String()
     expression = fields.String()
     dialect = fields.String()
+    skip_parsing = fields.Boolean(allow_none=True)
 
 
 class MFMetricType(str, Enum):
--- a/src/preset_cli/cli/superset/sync/dbt/metrics.py
+++ b/src/preset_cli/cli/superset/sync/dbt/metrics.py
@@ -8,11 +8,12 @@ This module is used to convert dbt metrics into Superset metrics.
 
 import json
 import logging
+import re
 from collections import defaultdict
 from typing import Dict, List, Optional, Set
 
 import sqlglot
-from sqlglot import Expression, exp, parse_one
+from sqlglot import Expression, ParseError, exp, parse_one
 from sqlglot.expressions import (
     Alias,
     Case,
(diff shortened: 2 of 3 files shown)


betodealmeida:
I'm confused. 🙂 When you say

> the {% raw %} macro also allows users to use Superset-specific Jinja in their metrics:

Is this something people actually want to do?

Why write this:

```yaml
metrics:
  - name: new_derived_again
    label: test
    calculation_method: derived
    expression: |
      {% raw %}
       (SUM(price_each)) * 1.2
      {% endraw %}
```

Why not write directly:

```yaml
metrics:
  - name: new_derived_again
    label: test
    calculation_method: derived
    expression: |
     (SUM(price_each)) * 1.2
```

?

Vitor-Avila:
@betodealmeida ohh maybe that was a bad example. Your example should also work (in the sense that dbt would compile properly), however the CLI wouldn't be able to sync it since the metric doesn't have any dependency.

The `{% raw %}` macro would be needed when using the Superset-specific Jinja macros:
``` yaml
metrics:
  - name: new_derived_again
    label: test
    calculation_method: derived
    expression: |
      {% raw %}
        SUM(
          {% for x in filter_values('x_values') %}
            {{ + x_values }}
          {% endfor %}
        )
      {% endraw %}
```

Issue #282, opened 2024-04-15
DBT metricflow error on standalone superset version


Santhin:
Packages version:
```  
preset-cli, version 0.2.14   
mf, version 0.205.0
```

Helm chart 
```
NAME            NAMESPACE       REVISION        UPDATED                                 STATUS          CHART           APP VERSION
superset        superset        3               2024-04-15 23:05:41.716187 +0200 CEST   deployed        superset-0.12.9 4.0.0   
```

Values
```
# Override defaults from https://github.com/apache/superset/blob/master/helm/superset/values.yaml

configOverrides:
  secret: |
    SECRET_KEY = ''
  # dbt model/column descriptions into Superset via PUT requests.
  disable_csrf: |
    WTF_CSRF_ENABLED = False

bootstrapScript: |
  #!/bin/bash
  pip3 install snowflake-sqlalchemy
  if [ ! -f ~/bootstrap ]; then echo "Running Superset with uid {{ .Values.runAsUser }}" > ~/bootstrap; fi

```

Example used for boostrapping https://github.com/dbt-labs/jaffle-sl-template/tree/main
Snowflake as a DWH 

Command that creates error
```
superset-cli -u admin -p admin http://localhost:8088 sync dbt ./target/manifest.json --profiles=profiles.yml --target=dev --project=ecommerce --import-db
```

Error log
```
Traceback (most recent call last):
  File "/Users/piotrsierkin/Documents/private/jaffle-sl-template/.venv/bin/superset-cli", line 8, in <module>
    sys.exit(superset_cli())
             ^^^^^^^^^^^^^^
  File "/Users/piotrsierkin/Documents/private/jaffle-sl-template/.venv/lib/python3.11/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/piotrsierkin/Documents/private/jaffle-sl-template/.venv/lib/python3.11/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/Users/piotrsierkin/Documents/private/jaffle-sl-template/.venv/lib/python3.11/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #283, opened 2024-04-16
chore: Pin sqlglot to <23

Vitor-Avila:
This PR sets an upper bound to the `sqlglot` version -- the CLI doesn't work properly when using a version `>=23.0.0`.
Fixes https://github.com/preset-io/backend-sdk/issues/282

Also bumped `pip-tools` as I was facing errors to run `pip-compile`.


betodealmeida (commented):
Why not set it to `>=23`, and update the call to `walk`? We're going to have to do that eventually.

Vitor-Avila:
@betodealmeida that's true and I'm happy to take a look on that as part of a bigger effort to review our dependencies (we might need to also bump `Jinja2` to be compatible with `metricflow` and other dbt dependencies).

betodealmeida (approved):

PR #284, opened 2024-04-16
chore(sync native): Attempt to get the DB connection UUIDs via API

Vitor-Avila:
During the `sync native` command, the CLI has to identify all existing DB connections in the target instance/Workspace (and their corresponding UUIDs). This is currently achieved by exporting each DB connection, and getting the `uuid` from the YAML file. 

This process was adopted because the `uuid` is not exposed in the API response for older Superset versions. However, it affects performance for two reasons:
* Each DB connection is exported (the more connections created, the longer it would take).
* There was a change in Superset that exporting a DB connection also includes all datasets, which can take time for connections with hundreds/thousands of datasets.

This PR defaults the process to get the `uuid` from the API response, and fallback in the old approach in case the API response doesn't have the `uuid` field.


betodealmeida (approved):
Ohhh, nice!

PR #285, opened 2024-04-19
fix(derived metrics): Support reusing metrics with Jinja in other metrics

Vitor-Avila:
https://github.com/preset-io/backend-sdk/pull/277 introduced the ability to sync metrics from dbt with Jinja syntax, which could also include other metrics. For example:

``` yaml
metrics:
  - name: revenue
    label: test label
    calculation_method: sum
    expression: price_each
  - name: derived_metric_with_jinja
    label: test label
    calculation_method: derived
    expression: |
      {% raw %}
        SUM(
          {% for x in filter_values('x_values') %}
            {{ + x_values }}
          {% endfor %}
        )
      {% endraw %}
      / {{ metric('revenue') }}
```

However, the other way around (creating a derived metric including another derived metric with Jinja) wasn't supported yet:
``` yaml
metrics:
  - name: derived_metric_with_jinja
    label: test label
    calculation_method: derived
    expression: |
      {% raw %}
        SUM(
          {% for x in filter_values('x_values') %}
            {{ + x_values }}
          {% endfor %}
        )
      {% endraw %}
  - name: revenue
    label: test label
    calculation_method: derived
    expression: {{ metric('derived_metric_with_jinja') }} / 2
```

This PR introduces support for this flow as well, and also improves the logic when the ``sqlglot`` parsing fails.

### DIFF
--- a/src/preset_cli/cli/superset/sync/dbt/metrics.py
+++ b/src/preset_cli/cli/superset/sync/dbt/metrics.py
@@ -89,31 +89,26 @@ def get_metric_expression(metric_name: str, metrics: Dict[str, MetricSchema]) ->
         return f"COUNT(DISTINCT {sql})"
 
     if type_ in {"expression", "derived"}:
+        if metric.get("skip_parsing"):
+            return remove_line_break_wrappers(sql)
+
         try:
             expression = sqlglot.parse_one(sql, dialect=metric["dialect"])
+            tokens = expression.find_all(exp.Column)
+
+            for token in tokens:
+                if token.sql() in metrics:
+                    parent_sql = get_metric_expression(token.sql(), metrics)
+                    parent_expression = sqlglot.parse_one(
+                        parent_sql,
+                        dialect=metric["dialect"],
+                    )
+                    token.replace(parent_expression)
+
+            return expression.sql(dialect=metric["dialect"])
         except ParseError:
-            for parent_metric in metric["depends_on"]:
-                parent_metric_name = parent_metric.split(".")[-1]
-                pattern = r"\b" + re.escape(parent_metric_name) + r"\b"
-                parent_metric_syntax = get_metric_expression(
-                    parent_metric_name,
-                    metrics,
-                )
-                sql = re.sub(pattern, parent_metric_syntax, sql)
+            sql = replace_metric_syntax(sql, metric["depends_on"], metrics)
             return sql
-
-        tokens = expression.find_all(exp.Column)
-
-        for token in tokens:
-            if token.sql() in metrics:
-                parent_sql = get_metric_expression(token.sql(), metrics)
-                parent_expression = sqlglot.parse_one(
-                    parent_sql,
-                    dialect=metric["dialect"],
-                )
-                token.replace(parent_expression)
-
-        return expression.sql(dialect=metric["dialect"])
 
     sorted_metric = dict(sorted(metric.items()))
     raise Exception(f"Unable to generate metric expression from: {sorted_metric}")
@@ -206,11 +201,7 @@ def get_metric_definition(
     kwargs = meta.pop("superset", {})
 
     return {
-        "expression": (
-            get_metric_expression(metric_name, metric_map)
-            if not metric.get("skip_parsing")
-            else metric.get("expression") or metric.get("sql")
-        ),
+        "expression": get_metric_expression(metric_name, metric_map),
         "metric_name": metric_name,
         "metric_type": (metric.get("type") or metric.get("calculation_method")),
         "verbose_name": metric.get("label", metric_name),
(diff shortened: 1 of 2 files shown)


src/preset_cli/cli/superset/sync/dbt/metrics.py
```diff
@@ -89,32 +89,27 @@ def get_metric_expression(metric_name: str, metrics: Dict[str, MetricSchema]) ->
         return f"COUNT(DISTINCT {sql})"
 
     if type_ in {"expression", "derived"}:
+        if metric.get("skip_parsing"):
+            return remove_line_break_wrappers(sql)
```
Vitor-Avila, line 93: Thought it made more sense to handle this `skip_parsing` scenario here (as part of the ``get_metric_expression`` method than to avoid calling it altogether. This is also useful to ensure that recursive calls to this method handle this logic properly.

src/preset_cli/cli/superset/sync/dbt/metrics.py
```diff
@@ -89,32 +89,27 @@ def get_metric_expression(metric_name: str, metrics: Dict[str, MetricSchema]) ->
...
+            for token in tokens:
+                if token.sql() in metrics:
+                    parent_sql = get_metric_expression(token.sql(), metrics)
+                    parent_expression = sqlglot.parse_one(
+                        parent_sql,
+                        dialect=metric["dialect"],
+                    )
+                    token.replace(parent_expression)
+
+            return expression.sql(dialect=metric["dialect"])
```
Vitor-Avila, line 108: Haven't really changed anything in this block, just moved it inside the `try/except` block as the `sqlglot.parse_one()` call in the `for` loop might also fail.

src/preset_cli/cli/superset/sync/dbt/metrics.py
```diff
@@ -399,3 +393,32 @@ def get_models_from_sql(
...
 
     return [model_map[ModelKey(table.db, table.name)] for table in sources]
+
+
+def remove_line_break_wrappers(sql: str) -> str:
+    """
+    Remove leading and trailing line breaks from SQL statements.
+    """
+    pattern = r"^\s*\n|\n\s*$"
+    return re.sub(pattern, "", sql, flags=re.MULTILINE)
```
Vitor-Avila, line 403: `sqlglot` handles this out of the box, but the fallback approach with `regex` doesn't.

src/preset_cli/cli/superset/sync/dbt/metrics.py
```diff
@@ -206,11 +201,7 @@ def get_metric_definition(
     kwargs = meta.pop("superset", {})
 
     return {
-        "expression": (
-            get_metric_expression(metric_name, metric_map)
-            if not metric.get("skip_parsing")
-            else metric.get("expression") or metric.get("sql")
```
Vitor-Avila, line 212: Logic that has been moved inside the `get_metric_expression`

src/preset_cli/cli/superset/sync/dbt/metrics.py
```diff
@@ -399,3 +393,32 @@ def get_models_from_sql(
...
+    This method is a fallback in case ``sqlglot`` raises a ``ParseError``.
+    """
+    for parent_metric in dependencies:
+        parent_metric_name = parent_metric.split(".")[-1]
+        pattern = r"\b" + re.escape(parent_metric_name) + r"\b"
+        parent_metric_syntax = get_metric_expression(
+            parent_metric_name,
+            metrics,
+        )
+        sql = re.sub(pattern, parent_metric_syntax, sql)
```
Vitor-Avila, line 422: Same syntax I previously had inside `get_metric_expression` (just moved to its own method).

src/preset_cli/cli/superset/sync/dbt/metrics.py
```diff
@@ -229,34 +220,37 @@ def get_superset_metrics_per_model(
...
-        if len(metric_models) == 0:
-            try:
-                metric_models.add(metric["meta"]["superset"].pop("model"))
+        # dbt supports creating derived metrics with raw syntax. In case the metric doesn't
+        # rely on other metrics (or rely on other metrics that aren't associated with any
+        # model), it's required to specify the dataset the metric should be associated with
+        # under the ``meta.superset.model`` key. If the derived metric is just an expression
+        # with no dependency, it's not required to parse the metric SQL.
+        if model := metric.get("meta", {}).get("superset", {}).pop("model", None):
+            if len(metric["depends_on"]) == 0:
```
adamcunnington-mlg, line 229: why not `if not metric["depends_on"]`?

src/preset_cli/cli/superset/sync/dbt/metrics.py
betodealmeida: Can't we do this with just `sql.strip()`?

betodealmeida (approved):

src/preset_cli/cli/superset/sync/dbt/metrics.py
Vitor-Avila: omg you're 100% right 🤦 talk about overcomplicating things lol lemme fix that real quick

PR #289, opened 2024-05-01
fix(dbt): Avoid raising an error when syncing specific models

Vitor-Avila:
The CLI allows users to sync specific models via dbt. As a consequence, when evaluating metrics it's expected that some will rely on models that aren't included in the ``model_map``. This PR changes the logic to avoid raising a  ``ValueError`` in this scenario and instead just ignore the metric.


betodealmeida (approved):
Nice!

PR #290, opened 2024-05-01
chore(dbt): Instruct users of missing dependency when creating a Snowflake virtual dataset

Vitor-Avila:
dbt is compatible with warehouses that support creating multiple DBs in the same instance (such as Snowflake). Since the `database` is specified in the DB connection string in Superset, when a model is created in a different database, the CLI creates it as a virtual dataset (using `select * from db_name.schema.table`). 

SQLAlchemy's `create_engine()` method might require the DB-specific driver to work. This PR adds proper instructions to users when facing this issue. It currently only covers `snowflake`, but covering more engines should be straightforward.

### DIFF
--- a/src/preset_cli/cli/superset/sync/dbt/lib.py
+++ b/src/preset_cli/cli/superset/sync/dbt/lib.py
@@ -210,8 +210,8 @@ def create_sqlalchemy_engine(url: URL):
     except NoSuchModuleError as exc:
         string_url = str(url)
         dialect = string_url.split("://", maxsplit=1)[0]
+        # TODO: Handle more DB engines that require an additional package
         if dialect == "snowflake":
-            # TODO: Check if we can progamatically read all options listed under options.extra_require
             raise CLIError(
                 (
                     "Missing required package. "
--- a/src/preset_cli/cli/superset/sync/dbt/lib.py
+++ b/src/preset_cli/cli/superset/sync/dbt/lib.py
@@ -12,9 +12,12 @@ from typing import Any, Dict, List, Optional, Tuple, TypedDict, Union
 
 import yaml
 from jinja2 import Environment
+from sqlalchemy.engine import create_engine
 from sqlalchemy.engine.url import URL
+from sqlalchemy.exc import NoSuchModuleError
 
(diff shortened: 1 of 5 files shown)


src/preset_cli/cli/superset/sync/dbt/lib.py
```diff
@@ -198,6 +201,31 @@ def build_snowflake_sqlalchemy_params(target: Dict[str, Any]) -> Dict[str, Any]:
...
+        dialect = string_url.split("://", maxsplit=1)[0]
+        # TODO: Handle more DB engines that require an additional package
+        if dialect == "snowflake":
+            raise CLIError(
+                (
+                    "Missing required package. "
+                    'Please run ``pip install "preset-cli[snowflake]"`` to install it.'
+                ),
+                1,
+            ) from exc
```
Vitor-Avila, line 221: In the future this could be something like:
``` python
if dialect in {"snowflake", "redshift", "bigquery"}:
    raise CLIError(
        (
            "Missing required package. "
            f'Please run ``pip install "preset-cli[{dialect}]"`` to install it.'
        ),
        1,
    ) from exc
```

src/preset_cli/cli/superset/sync/dbt/lib.py
```diff
@@ -198,6 +201,31 @@ def build_snowflake_sqlalchemy_params(target: Dict[str, Any]) -> Dict[str, Any]:
     return parameters
 
 
+def create_sqlalchemy_engine(url: URL):
```
betodealmeida, line 204:
```suggestion
def create_sqlalchemy_engine(url: URL) -> Engine:
```

With:

```python
from sqlalchemy.engine import Engine
```

Also, maybe we should rename this to `create_engine_with_check` or something more explicit? To make it clear how it's different from `create_engine`.

betodealmeida (approved):
Looks great! Left a few comments.

src/preset_cli/cli/superset/sync/dbt/lib.py
Vitor-Avila: great catch! and great idea!

PR #291, opened 2024-05-09
fix(dbt): Remove Redshift-specific column metadata

Vitor-Avila:
Redshift reflection query returns additional metadata that isn't compatible with the Superset column schema. This PR removes the `info` key when syncing columns.

I think long-term solution is to enforce the Superset schema for columns and metrics accordingly -- I'll look into that in a future PR.


betodealmeida (approved):
Yeah, let's add a Marshmallow schema that ignores unknown fields and use it here instead (in the future).

PR #292, opened 2024-05-09
fix(import-assets): Include all dashboard depedencies when using the the --split flag

Vitor-Avila:
The `sync native` / `import-assets` command allows users to include a `--split` flag to import resources individually. It's possible to create dashboard filters using datasets that aren't used by any chart. The logic was currently not including these datasets when importing the dashboard, resulting in an import error.

### DIFF
--- a/src/preset_cli/cli/superset/sync/native/command.py
+++ b/src/preset_cli/cli/superset/sync/native/command.py
@@ -11,7 +11,7 @@ from datetime import datetime, timezone
 from io import BytesIO
 from pathlib import Path
 from types import ModuleType
-from typing import Any, Dict, Iterator, Tuple
+from typing import Any, Dict, Iterator, Set, Tuple
 from zipfile import ZipFile
 
 import backoff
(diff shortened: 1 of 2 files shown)


src/preset_cli/cli/superset/sync/native/command.py
```diff
@@ -310,6 +320,19 @@ def get_charts_uuids(config: AssetConfig) -> Iterator[str]:
...
 
 
+def get_dataset_filter_uuids(config: AssetConfig) -> Set[str]:
+    """
+    Extract dataset UUID for datasets that are used in dashboard filters.
+    """
+    dataset_uuids = set()
+    for filter_config in config["metadata"].get("native_filter_configuration", []):
+        for target in filter_config["targets"]:
+            if uuid := target.get("datasetUuid"):
```
Vitor-Avila, line 330: From my tests:
* Even if the dashboard doesn't have any elements, it always have the `position` and `metadata` keys/configurations.
* If the dashboard never had a filter created, it might not include the `native_filter_configuration` key.
* Even filters that don't have a target (such as temporal) still include a blank `targets` key.
* If the filter dataset was deleted, its `target` config won't have a `datasetUuid` field.

eschutho (approved):

PR #295, opened 2024-05-10
chore: avoid virtual datasets whenever possible

betodealmeida:
With SIP-95 implemented we don't need virtual datasets anymore, yay!

Note: ideally this still requires testing with a real project, and also for the SIP-95 branch to be deployed at Preset.

### DIFF
--- a/src/preset_cli/cli/superset/sync/dbt/datasets.py
+++ b/src/preset_cli/cli/superset/sync/dbt/datasets.py
@@ -82,15 +82,10 @@ def create_dataset(
     except SupersetError as ex:
         if not no_catalog_support(ex):
             raise ex
+        del kwargs["catalog"]
 
     url = make_url(database["sqlalchemy_uri"])
-    if model_in_database(model, url):
-        kwargs = {
-            "database": database["id"],
-            "schema": model["schema"],
-            "table_name": model.get("alias") or model["name"],
-        }
-    else:
+    if not model_in_database(model, url):
         engine = create_engine_with_check(url)
         quote = engine.dialect.identifier_preparer.quote
         source = ".".join(quote(model[key]) for key in ("database", "schema", "name"))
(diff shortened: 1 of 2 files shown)


src/preset_cli/cli/superset/sync/dbt/datasets.py
```diff
@@ -91,6 +104,32 @@ def create_dataset(
...
+    """
+    for error in ex.errors:
+        try:
+            message = json.loads(error["message"])
+            if "Unknown field." in message["message"]["catalog"]:
+                return True
+        except Exception:  # pylint: disable=broad-except
+            pass
+
+    return False
```
Vitor-Avila, line 130: Nice!

src/preset_cli/cli/superset/sync/dbt/datasets.py
```diff
@@ -70,6 +70,19 @@ def create_dataset(
...
+        "schema": model["schema"],
+        "table_name": model.get("alias") or model["name"],
+    }
+    try:
+        # try to create dataset with catalog
+        return client.create_dataset(**kwargs)
+    except SupersetError as ex:
+        if not no_catalog_support(ex):
+            raise ex
+
```
Vitor-Avila, line 85: You could just delete the `catalog` key here (`del kwargs["catalog"]`)...

src/preset_cli/cli/superset/sync/dbt/datasets.py
```diff
@@ -70,6 +70,19 @@ def create_dataset(
...
     url = make_url(database["sqlalchemy_uri"])
     if model_in_database(model, url):
```
Vitor-Avila, line 87: ... then avoid this `if/else` and have a single `if not model_in_database(model, url)`

Vitor-Avila (approved):
left a non-blocking suggestion

PR #299, opened 2024-06-03
chore(dbt sync): Deprecate using the dbt_project file to trigger a dbt sync

Vitor-Avila:
It's possible to trigger the dbt Core sync specifying the path to either the `manifest.json` file or to the `dbt_project.yml` file. The former would currently display a `DeprecationWarning`.

[dbt deprecated the `target-path` as a config in the `dbt_project.yml`](https://docs.getdbt.com/reference/global-configs/about-global-configs) (which is used when triggering the command with this file) so this PR is moving the deprecated warning to when syncing with the path to `dbt_project.yml` file specified. For simplicity, I would recommend dropping support entirely in a future release and relying solely in the `manifest.json` file.

This PR also moves away from using `warnings.warn()` as it breaks logging (not a concern when using the `preset-cli` as a standalone CLI, but could be impactful when using it as a dependency and relying on its methods).

### DIFF
--- a/src/preset_cli/cli/superset/sync/dbt/command.py
+++ b/src/preset_cli/cli/superset/sync/dbt/command.py
@@ -36,7 +36,7 @@ from preset_cli.cli.superset.sync.dbt.metrics import (
     get_superset_metrics_per_model,
 )
 from preset_cli.exceptions import CLIError, DatabaseNotFoundError
-from preset_cli.lib import log_warning, raise_cli_errors
+from preset_cli.lib import raise_cli_errors
 
 _logger = logging.getLogger(__name__)
 
@@ -137,8 +137,6 @@ def dbt_core(  # pylint: disable=too-many-arguments, too-many-branches, too-many
     url = URL(ctx.obj["INSTANCE"])
     client = SupersetClient(url, auth)
     deprecation_notice: bool = False
-    if raise_failures:
-        warnings.simplefilter("always", DeprecationWarning)
 
     if (preserve_columns or preserve_metadata) and merge_metadata:
         error_message = (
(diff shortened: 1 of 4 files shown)


src/preset_cli/cli/superset/sync/dbt/command.py
```diff
@@ -160,17 +157,18 @@ def dbt_core(  # pylint: disable=too-many-arguments, too-many-branches, too-many
             "The managed externally feature was only introduced in Superset v1.5."
             "Make sure you are running a compatible version."
         )
-        log_warning(warn_message, UserWarning)
+        _logger.debug(warn_message)
```
Vitor-Avila, line 160: Changed it to `debug` as it might make more sense (given it might not be a concern).

src/preset_cli/cli/superset/sync/dbt/command.py
```diff
@@ -160,17 +157,18 @@ def dbt_core(  # pylint: disable=too-many-arguments, too-many-branches, too-many
...
-        )
-        log_warning(warn_message, DeprecationWarning)
         manifest = file_path
         profile = project = project or "default"
     elif file_path.name == "dbt_project.yml":
+        deprecation_notice = True
+        warn_message = (
+            "Passing the dbt_project.yml file is deprecated and "
+            "will be removed in a future version. "
+            "Please pass the manifest.json file instead."
```
betodealmeida, line 169: lol

betodealmeida (approved):

src/preset_cli/cli/superset/sync/dbt/command.py
Vitor-Avila: Hoping to have a single supported method soon.

PR #301, opened 2024-06-13
fix(dbt Cloud): Support syncing a project without a semantic layer

Vitor-Avila:
If the project doesn't have a semantic layer, the query to fetch semantic layer metrics return below response:
``` json
{
    "data": null,
    "errors": [
        {
            "message": "Empty semantic manifest was found. Ensure that you have semantic models defined.",
            "locations": [{"line": 3, "column": 17}],
            "path": ["metrics"],
        },
    ],
}
```

This PR adds support for this scenario.


betodealmeida (approved):
Thanks!

Issue #288, opened 2024-04-29
sync dbt fails on Athena from version 0.2.12 


advincze-hivemq:
to provision our models that were built by dbt on AWS Athena we run the following command:

preset-cli superset sync dbt target/manifest.json  --project=<project> --target=prod --disallow-edits --select <model>

in version 0.2.11 everything is ok, 
from version 0.2.12 to the current one we get the following error message: 
```
Traceback (most recent call last):
  File "/Users/adam/src/github.com/hivemq/dwh-core/.venv/bin/preset-cli", line 8, in <module>
    sys.exit(preset_cli())
  File "/Users/adam/src/github.com/hivemq/dwh-core/.venv/lib/python3.10/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "/Users/adam/src/github.com/hivemq/dwh-core/.venv/lib/python3.10/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "/Users/adam/src/github.com/hivemq/dwh-core/.venv/lib/python3.10/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/adam/src/github.com/hivemq/dwh-core/.venv/lib/python3.10/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/adam/src/github.com/hivemq/dwh-core/.venv/lib/python3.10/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/adam/src/github.com/hivemq/dwh-core/.venv/lib/python3.10/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/adam/src/github.com/hivemq/dwh-core/.venv/lib/python3.10/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "/Users/adam/src/github.com/hivemq/dwh-core/.venv/lib/python3.10/site-packages/click/decorators.py", line 33, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/Users/adam/src/github.com/hivemq/dwh-core/.venv/lib/python3.10/site-packages/preset_cli/cli/superset/main.py", line 117, in new_command
    ctx.invoke(command, *args, **kwargs)
[...]

PR #302, opened 2024-06-20
fix(dbt Core): Support syncing metrics without a MetricFlow dialect

Vitor-Avila:
MetricFlow is not compatible with all DB engines supported by dbt Core yet. Currently, triggering a dbt Core sync in this scenario would raise an error. This PR gracefully handles this case and still processes legacy metrics.

Fixes https://github.com/preset-io/backend-sdk/issues/288.

### DIFF
--- a/src/preset_cli/cli/superset/sync/dbt/command.py
+++ b/src/preset_cli/cli/superset/sync/dbt/command.py
@@ -111,7 +111,7 @@ _logger = logging.getLogger(__name__)
 )
 @raise_cli_errors
 @click.pass_context
-def dbt_core(  # pylint: disable=too-many-arguments, too-many-branches, too-many-locals ,too-many-statements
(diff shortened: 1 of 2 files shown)


src/preset_cli/cli/superset/sync/dbt/command.py
```diff
@@ -111,7 +111,7 @@
 )
 @raise_cli_errors
 @click.pass_context
-def dbt_core(  # pylint: disable=too-many-arguments, too-many-branches, too-many-locals ,too-many-statements
+def dbt_core(  # pylint: disable=too-many-arguments, too-many-branches, too-many-locals ,too-many-statements # noqa: C901
```
Vitor-Avila, line 114: There's some logic getting deprecated in this method that should allow us to remove `# noqa: C901` soon. I could also plan a bigger refactor to break this down into smaller methods.

eschutho (approved):

PR #306, opened 2024-06-26
fix(dbt Cloud): Update dbt metadata GraphQL query

Vitor-Avila:
We were previously fetching legacy metrics with the `metrics(jobId: $jobId)` query, which was not working properly with new jobs as the `jobId` is an `Int!` for this query (and job IDs are exceeding the 32-bit limit).

This PR uses the `job(id: $jobId)` query which accepts/support a `BigInt!`.


betodealmeida (approved):
loveit

PR #312, opened 2024-08-14
fix(pre-commit): Updating mirrors-mypy dependencies

Vitor-Avila:
[types-pkg-resources](https://pypi.org/project/types-pkg-resources) have been yanked -- they recommend using `types-setuptools` instead.


betodealmeida (approved):
Oh, I had to do this in shillelagh as well.

PR #313, opened 2024-08-19
fix(dbt): Keep syncing datasets if one failed to sync columns

Vitor-Avila:
When syncing datasets with the `--preserve-metadata` or `--merge-metadata` flags, a request to sync the dataset columns is triggered. Currently, in case this request fails the command execution is aborted. 

This PR gracefully handles failures, adding the failed dataset to the list of `failed_datasets` (which can be listed at the end in case `--raise-failures` is used).

### DIFF
--- a/src/preset_cli/cli/superset/sync/dbt/datasets.py
+++ b/src/preset_cli/cli/superset/sync/dbt/datasets.py
@@ -349,11 +349,17 @@ def sync_datasets(  # pylint: disable=too-many-locals, too-many-arguments
         # compute columns
         final_dataset_columns = []
         if not reload_columns:
-            refreshed_columns_list = client.get_refreshed_dataset_columns(dataset["id"])
-            final_dataset_columns = compute_columns(
-                dataset["columns"],
-                refreshed_columns_list,
-            )
+            try:
+                refreshed_columns_list = client.get_refreshed_dataset_columns(
+                    dataset["id"],
+                )
+                final_dataset_columns = compute_columns(
+                    dataset["columns"],
+                    refreshed_columns_list,
+                )
+            except SupersetError:
+                failed_datasets.append(model["unique_id"])
+                continue
 
         # compute update payload
         update = compute_dataset_metadata(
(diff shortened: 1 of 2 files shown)


Antonio-RiveroMartnez (approved):

src/preset_cli/cli/superset/sync/dbt/datasets.py
```diff
@@ -349,11 +349,17 @@ def sync_datasets(  # pylint: disable=too-many-locals, too-many-arguments
...
+            try:
+                refreshed_columns_list = client.get_refreshed_dataset_columns(
+                    dataset["id"],
+                )
+                final_dataset_columns = compute_columns(
+                    dataset["columns"],
+                    refreshed_columns_list,
+                )
+            except SupersetError:
+                failed_datasets.append(model["unique_id"])
```
Antonio-RiveroMartnez, line 361: Not very familiar with this part of the code, but since this list is visible only using a certain flag (`--raise-failures`) it would be good to put that info somewhere in the help of the cli method if not there already.

Vitor-Avila: thanks for the feedback, @Antonio-RiveroMartnez 🙏 I can confirm the flag is documented in the ``help`` command:
``` bash
(local-cli) ➜  backend-sdk git:(dbt-cloud-versionless) preset-cli superset sync dbt-core --help
Usage: preset-cli superset sync dbt-core [OPTIONS] FILE

Options:
  --project TEXT              Name of the dbt project
  --target TEXT               Target name
  --profiles PATH             Location of profiles.yml file
  --exposures PATH            Path to file where exposures will be written
  --import-db                 Import (or update) the database connection to
                              Superset
  --disallow-edits            Mark resources as managed externally to prevent
                              edits
  --external-url-prefix TEXT  Base URL for resources
  -s, --select TEXT           Model selection
  -x, --exclude TEXT          Models to exclude
  --exposures-only            Do not sync models to datasets and only fetch
                              exposures instead
  --preserve-columns          Preserve column and metric configurations
                              defined in Preset
  --preserve-metadata         Preserve column and metric configurations
                              defined in Preset
  --merge-metadata            Update Preset configurations based on dbt
                              metadata. Preset-only metrics are preserved
  --raise-failures            End the execution with an error if a model fails
                              to sync or a deprecated feature is used
  --help                      Show this message and exit.
```

Since this is a customer-facing CLI, we also have this logic documented [here](https://docs.preset.io/docs/dbt-sync).

PR #316, opened 2024-08-29
Add option to force Unix EOL on YAML export via the CLI

kevmk04:
PR adds an optional argument to the CLI export commands to force Unix EOLs for the exported files.  This is intended to address issues when exporting into Windows environments where we have multiple developers using different OS.

### DIFF
--- a/src/preset_cli/cli/superset/export.py
+++ b/src/preset_cli/cli/superset/export.py
@@ -21,6 +21,11 @@ JINJA2_CLOSE_MARKER = "__JINJA2_CLOSE__"
 assert JINJA2_OPEN_MARKER != JINJA2_CLOSE_MARKER
 
 
+def get_newline_char(force_unix_eol: bool = False) -> str | None:
+    """Returns the newline character used by the open function"""
+    return "\n" if force_unix_eol else None
+
+
 @click.command()
 @click.argument("directory", type=click.Path(exists=True, resolve_path=True))
 @click.option(
@@ -158,10 +163,7 @@ def export_resource(  # pylint: disable=too-many-arguments, too-many-locals
 
             file_contents = yaml.dump(asset_content, sort_keys=False)
 
-        if force_unix_eol:
-            newline = "\n"
-        else:
-            newline = None
+        newline = get_newline_char(force_unix_eol)
         with open(target, "w", encoding="utf-8", newline=newline) as output:
             output.write(file_contents)
 
@@ -257,10 +259,7 @@ def export_users(
         {k: v for k, v in user.items() if k != "id"} for user in client.export_users()
     ]
 
-    if force_unix_eol:
-        newline = "\n"
-    else:
-        newline = None
+    newline = get_newline_char(force_unix_eol)
     with open(path, "w", encoding="utf-8", newline=newline) as output:
         yaml.dump(users, output)
 
@@ -290,10 +289,7 @@ def export_roles(
     url = URL(ctx.obj["INSTANCE"])
     client = SupersetClient(url, auth)
 
-    if force_unix_eol:
-        newline = "\n"
-    else:
-        newline = None
+    newline = get_newline_char(force_unix_eol)
     with open(path, "w", encoding="utf-8", newline=newline) as output:
         yaml.dump(list(client.export_roles()), output)
 
@@ -323,10 +319,7 @@ def export_rls(
     url = URL(ctx.obj["INSTANCE"])
     client = SupersetClient(url, auth)
 
-    if force_unix_eol:
-        newline = "\n"
-    else:
-        newline = None
+    newline = get_newline_char(force_unix_eol)
     with open(path, "w", encoding="utf-8", newline=newline) as output:
         yaml.dump(list(client.export_rls()), output, sort_keys=False)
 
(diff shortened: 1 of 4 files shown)


kevmk04:
I believe this is ready for review!

src/preset_cli/cli/superset/export.py
```diff
@@ -149,7 +158,11 @@ def export_resource(  # pylint: disable=too-many-arguments, too-many-locals
 
             file_contents = yaml.dump(asset_content, sort_keys=False)
 
-        with open(target, "w", encoding="utf-8") as output:
+        if force_unix_eol:
+            newline = "\n"
+        else:
+            newline = None
```
Vitor-Avila, line 164: small nit:
``` python
newline = "\n" if force_unix_eol else None
```

src/preset_cli/cli/superset/export.py
```diff
@@ -244,16 +271,30 @@ def export_users(ctx: click.core.Context, path: str) -> None:
...
     """
     auth = ctx.obj["AUTH"]
     url = URL(ctx.obj["INSTANCE"])
     client = SupersetClient(url, auth)
 
-    with open(path, "w", encoding="utf-8") as output:
+    if force_unix_eol:
+        newline = "\n"
+    else:
+        newline = None
```
Vitor-Avila, line 296: Since this happens in a few different places, another potential approach is to create a small helper function in this file to evaluate the `force_unix_eol` value and return the appropriate `newline` configuration, and then in each of these occurrences you would only call this helper method passing the `force_unix_eol` value.

What are your thoughts?

Vitor-Avila:
thank you so much for the contribution, @kevmk04! 🙏 This is super cool. 🙌 
Left some minor comments -- let me know your thoughts.

src/preset_cli/cli/superset/export.py
kevmk04: Yes I think a helper method makes sense. Latest changes include this.

kevmk04:
@Vitor-Avila Should be all good now with your PR comments.

tests/cli/superset/export_test.py
```diff
@@ -509,6 +526,52 @@ def test_export_users(mocker: MockerFixture, fs: FakeFilesystem) -> None:
     ]
 
 
+def test_export_users_force_unix_eol_enable(
+    mocker: MockerFixture,
+    fs: FakeFilesystem,
+) -> None:
+    """
+    Test the ``export_users`` command.
```
Vitor-Avila, line 534: Could we add to the docstring that the test is with `force_unix_eol` set to `True`? Something like:
```
Test the ``export_users`` command with ``--force-unix-eol`` flag.
```

tests/cli/superset/export_test.py
```diff
@@ -541,6 +604,46 @@ def test_export_roles(mocker: MockerFixture, fs: FakeFilesystem) -> None:
     ]
 
 
+def test_export_roles_force_unix_eol_enable(
+    mocker: MockerFixture,
+    fs: FakeFilesystem,
+) -> None:
+    """
+    Test the ``export_roles`` command.
```
Vitor-Avila, line 612: Same thing here.

tests/cli/superset/export_test.py
```diff
@@ -583,6 +686,51 @@ def test_export_rls(mocker: MockerFixture, fs: FakeFilesystem) -> None:
     ]
 
 
+def test_export_rls_force_unix_eol_enable(
+    mocker: MockerFixture,
+    fs: FakeFilesystem,
+) -> None:
+    """
+    Test the ``export_rls`` command.
```
Vitor-Avila, line 694: Here too

tests/cli/superset/export_test.py
```diff
@@ -623,6 +771,49 @@ def test_export_ownership(mocker: MockerFixture, fs: FakeFilesystem) -> None:
     }
 
 
+def test_export_ownership_force_unix_eol_enable(
+    mocker: MockerFixture,
+    fs: FakeFilesystem,
+) -> None:
+    """
+    Test the ``export_ownership`` command.
```
Vitor-Avila, line 779: This one as well

tests/cli/superset/export_test.py
```diff
@@ -730,6 +925,124 @@ def test_export_resource_jinja_escaping_disabled_command(
...
+    WHERE action is null
+    {% for action in filter_values('action_type') %}
+        or action = '{{ action }}'
+    {% endfor %}
+{% endif %}
+GROUP BY action""",
+        }
+
+    # metadata file should be ignored
+    assert not (root / "metadata.yaml").exists()
```
Vitor-Avila, line 974: We can update the docstring here as well, and I think this test is still evaluating the Jinja behavior. Maybe something like this would be more appropriate?
``` python
root = Path("/path/to/root")
fs.create_dir(root)

client = mocker.MagicMock()
client.export_zip.return_value = chart_export
get_newline_char = mocker.patch("preset_cli.cli.superset.export.get_newline_char")
get_newline_char.return_value = "\n"

export_resource(
    resource_name="dataset",
    requested_ids=set(),
    root=root,
    client=client,
    overwrite=False,
    disable_jinja_escaping=False,
    force_unix_eol=True,
)

get_newline_char.assert_called_once_with(True)
```

Vitor-Avila (approved):
Thank you so much for working on this, @kevmk04! I left some minor comments around the tests. We should be good to merge this once they're addressed. 🙌

tests/cli/superset/export_test.py
kevmk04: Yeah makes sense, easier with the helper function.

PR #319, opened 2024-09-17
Fix sqlglot dependency conflict with superset 4 and fix superset auth/reauth to use only the API

joaopamaral:
There is a conflict with superset 4 dependencies and using only the API to authenticate it's better than using web scraping tool to avoid unexpected issues when the UI changes.


betodealmeida:
Nice! I'll take a look Monday. 

One thing to keep in mind is that we want to always keep the CLI backwards compatible, so ideally we want to fallback scraping if the CSRF endpoint is not available. But it seems to be pretty old, so maybe it's OK in this case.

Vitor-Avila:
I haven't tested this, but just a heads up that the `sqlglot` current upper bound was added to address a bug: https://github.com/preset-io/backend-sdk/pull/283 so we might need to double check if this newer limit would require any modifications to the dbt logic.

joaopamaral:
Hi @Vitor-Avila! Sorry, I've missed your comment.

I've bumped the `sqlglot` to use the same version of the latest superset and fixed some expression loops for the new version (now only `tests/cli/superset/sync/dbt/metrics_test.py::test_replace_metric_syntax` test is failing). As you said, the jinja template bug for dbt queries remains. I'll try to detect the block start `{{` and block end `}}` and make the tokens between as a placeholder to allow it to run without changing the content. The `{%` `%}` is correctly identified as block start/end but the content between is defined as a variable and not placeholder (so it's failing to parse).

> I've noticed that for this test we have a ParseError in the past (captured in the exception), but in the new `sqlglot` new version it parse both {{ }} as `STRUCT(STRUCT())` and not raise this exception.

joaopamaral:
I've added the `replace_jinja_tokens` to make sqlglot identify `{{`/`}}` it as a block start/end.

https://github.com/preset-io/backend-sdk/blob/86f5ad6e750ff0d5b97c107249d1fade3873aa2b/src/preset_cli/cli/superset/sync/dbt/metrics.py#L126-L135

All tests are passing now.

setup.cfg
```diff
@@ -4,7 +4,7 @@
 # https://setuptools.pypa.io/en/latest/references/keywords.html
 
 [metadata]
-name = preset-cli
+name = preset-cli-a8c
```
Vitor-Avila, line 7: Was this intentional?

src/preset_cli/auth/main.py
```diff
@@ -46,6 +48,8 @@ def reauth(self, r: Response, *args: Any, **kwargs: Any) -> Response:
         if r.status_code != 401:
             return r
 
+        _logger.info("Token expired. Re-authenticating...")
```
Vitor-Avila, line 51: Could this be a `DEBUG` instead? I think it's useful when debugging but we might not need it logged for all executions.

src/preset_cli/auth/superset.py
```diff
@@ -28,24 +27,43 @@ def __init__(self, baseurl: URL, username: str, password: Optional[str] = None):
...
         return {"X-CSRFToken": self.csrf_token} if self.csrf_token else {}
 
+    def get_access_token(self):
+        """
+        Get an access token from superset API: api/v1/security/login.
+        """
+        body = {
+            "username": self.username,
+            "password": self.password,
+            "provider": "ldap",
```
Vitor-Avila, line 37: Would `ldap` work in all cases? I'm not too familiar with the provider variances in Superset, but I know that on my end when testing I set it to `db`. Wondering if this should be a dynamic config.

setup.cfg
joaopamaral: Sorry, it was left during the tests. I've removed it.

src/preset_cli/auth/main.py
joaopamaral: Sure, changed to debug.

Vitor-Avila:
Hey @joaopamaral,

Checking this PR, I can see the following changes:
* `sqlglot` bump (and required changes)
* Changes to the Superset client
* Changes to the Superset auth flow
* Changes to the `Makefile`

I left some first pass comments, but I'm wondering if we should actually split this work into individual PRs? This would allow us to merge them individually as we make progress.

It seems that the CLA signature also didn't reflect properly FYI.

src/preset_cli/auth/superset.py
joaopamaral: Good point. I think it's better to change to dynamic config. Let me change it.

setup.cfg
Vitor-Avila: no worries, and thank you! 🙏

src/preset_cli/auth/main.py
Vitor-Avila: Thanks!

joaopamaral:
Hi @Vitor-Avila! Thanks for the review!
Sure we can split into more PRs!
I'll work on it during the weekend.

Vitor-Avila:
No rush @joaopamaral. And thank YOU for these contributions. We really appreciate it 🙌

PR #320, opened 2024-10-10
chore: Paginate Manager API calls

Vitor-Avila:
Manager APIs now support pagination. This PR implements pagination to Manager API calls.

### DIFF
--- a/src/preset_cli/api/clients/preset.py
+++ b/src/preset_cli/api/clients/preset.py
@@ -67,9 +67,7 @@ class PresetClient:  # pylint: disable=too-few-public-methods
         page_number = 1
         while True:
             params = {"page_number": page_number, "page_size": MANAGER_MAX_PAGE_SIZE}
-            url = (
-                self.get_base_url() / "teams" / team_name / "memberships" / "" % params
-            )
+            url = self.get_base_url() / "teams" / team_name / "memberships" % params
             _logger.debug("GET %s", url)
             response = self.session.get(url)
             validate_response(response)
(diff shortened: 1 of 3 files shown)


src/preset_cli/api/clients/preset.py
```diff
@@ -61,13 +63,24 @@ def get_team_members(self, team_name: str) -> List[Any]:
...
-        url = self.get_base_url() / "teams" / team_name / "memberships"
-        _logger.debug("GET %s", url)
-        response = self.session.get(url)
-        validate_response(response)
+        users = []
+        page_number = 1
+        while True:
+            params = {"page_number": page_number, "page_size": MANAGER_MAX_PAGE_SIZE}
+            url = (
+                self.get_base_url() / "teams" / team_name / "memberships" / "" % params
```
betodealmeida, line 71: Oh, do we need the trailing slash here, before `params`?

src/preset_cli/api/clients/preset.py
```diff
@@ -123,27 +136,42 @@ def export_users(self, workspace_url: URL) -> Iterator[UserType]:
...
+        while True:
+            params = {"page_number": page_number, "page_size": MANAGER_MAX_PAGE_SIZE}
+            url = (
+                self.get_base_url()
+                / "teams"
+                / team_name
+                / "workspaces"
+                / str(workspace_id)
+                / "memberships"
+                / ""
```
betodealmeida, line 150: Ditto here.

betodealmeida (approved):

src/preset_cli/api/clients/preset.py
Vitor-Avila: We don't, actually. Both would work -- lemme remove them.

PR #322, opened 2024-10-15
feat: Support importing specific asset types

Vitor-Avila:
By default, all import operations are executed using the Assets endpoint (`/api/v1/assets/import`), which overwrites all content in the target instance. While this is extremely useful when managing assets as code (to enforce the repo is always the source of truth) there are certain use-cases in which you might want to push assets without overwriting dependencies (updating a dataset without affecting the DB connection, push a chart update without updating datasets, etc) similar to the UI import flow.

This PR adds this capability to the `sync native` command. It's now possible to specify an ``--asset-type`` in the command (`dashboard` | `chart` | `dataset` | `database`) which defines the API endpoint used by the import (`/api/v1/{{asset_type}}/import`). As a consequence:

* When `--asset-type` is set to `database`:
     * The DB connections will get created/overwritten.
     * New datasets will get created. Existing datasets won't get overwritten.
     * Other asset types are ignored.

* When `--asset-type` is set to `dataset`:
     * The dataset(s) will get created/overwritten.
     * New DB connections will get created, existing ones won't be overwritten.
     * Other asset types are ignored.

* When `--asset-type` is set to `chart`:
     * The chart(s) will get created/overwritten.
     * New datasets will get created, existing ones won't be overwritten.
     * New DB connections will get created, existing ones won't be overwritten.
     * Dashboards would be ignored.

* When `--asset-type` is set to `dashboard`:
     * The dashboard(s) will get created/overwritten.
     * New datasets will get created, existing ones won't be overwritten.
     * New charts will get created, existing ones won't be overwritten.
     * New DB connections will get created, existing ones won't be overwritten.

Heads up that this behavior is not incorporated in the CLI, but instead the behavior for these API endpoints.

### DIFF
--- a/src/preset_cli/cli/superset/sync/native/command.py
+++ b/src/preset_cli/cli/superset/sync/native/command.py
@@ -8,10 +8,11 @@ import json
 import logging
 import os
 from datetime import datetime, timezone
+from enum import Enum
 from io import BytesIO
 from pathlib import Path
 from types import ModuleType
-from typing import Any, Dict, Iterator, Set, Tuple
+from typing import Any, Dict, Iterator, Optional, Set, Tuple
 from zipfile import ZipFile
 
 import backoff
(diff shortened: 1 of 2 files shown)


betodealmeida (approved):
Oh, nice! This is a very clean solution! ❤️

PR #325, opened 2024-11-06
feat(dbt): Support defining default values for columns and metrics

Vitor-Avila:
This PR adds a new feature for the dbt sync: the ability to specify default configs for metrics and columns. This allows you to have default values at the model level, and then if needed overwrite these for specific columns and metrics.

To re-cap, there are 3 dbt sync behaviors:

### Default (`reload_columns` is set to `True` and `merge_metadata` is set to `False`)
All metrics defined in dbt are synced to Superset (if the metric already exists in Superset, it gets entirely replaced by dbt's metadata). Column definitions from dbt are also synced to Superset.

In this flow, metric default configs would be applied to all metrics (unless the config is defined at the metric level), and column default configs would be applied to all columns (unless the config is defined at the column level) - even columns that are not defined in dbt.

### Merge metadata (`reload_columns` is set to `False` and `merge_metadata` is set to `True`)
All metrics defined in dbt are synced to Superset (if the metric already exists in Superset, it gets entirely replaced by dbt's metadata). Column definitions from dbt are also synced to Superset. 

In this flow, metric default configs would be applied to all metrics (even the ones that only exist in Superset), and column default configs would be applied to all columns (even columns not defined in dbt).

### Preserve metadata (both `reload_columns and `merge_metadata` are set to `False`)
In this flow, Superset is considered the source of truth. Metrics are only synced from dbt if they don't exist in Superset yet. Column definitions are only synced from dbt if not defined in Superset yet.

Here, metric default configs are only applied to dbt metrics that are getting synced (new metrics),  and to columns that are defined in dbt and getting synced.

### DIFF
--- a/src/preset_cli/cli/superset/sync/dbt/datasets.py
+++ b/src/preset_cli/cli/superset/sync/dbt/datasets.py
@@ -273,7 +273,7 @@ def compute_columns_metadata(
             for key, value in dbt_metadata[name].items():
                 if reload_columns or merge_metadata or not column.get(key):
                     column[key] = value
-        elif column_defaults:
+        elif column_defaults and (reload_columns or merge_metadata):
             for key, value in column_defaults.items():
                 column[key] = value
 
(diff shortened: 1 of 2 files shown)


betodealmeida (approved):
Looks good!

PR #328, opened 2024-12-12
feat(sync native): Continue import with split when an error happens

Vitor-Avila:
When the `sync native` command is used with the `--split` flag, it creates a `checkpoint.log` file to keep track of the import progress, so that if a failure happens, the next import attempt continues from where it's left.

This PR introduces a new optional flag that can be used with `--split`: `--continue-on-error`. When these two flags are combined, the behavior is different:
* Only asses that fail to import are added to the log file.
* The log file is named `failures.log`.
* If a failure happens, the import process continues. 

This is helpful when migrating from a Superset instance that contains some orphaned assets (charts linked to deleted datasets, etc) which would produce a lot of failures. This allows the process continue.

### DIFF
--- a/src/preset_cli/cli/superset/sync/native/command.py
+++ b/src/preset_cli/cli/superset/sync/native/command.py
@@ -2,6 +2,8 @@
 A command to sync Superset exports into a Superset instance.
 """
 
+from __future__ import annotations
+
 import getpass
 import importlib.util
 import json
@@ -202,6 +204,13 @@ def render_yaml(path: Path, env: Dict[str, Any]) -> Dict[str, Any]:
     help="Split imports into individual assets",
 )
 @click.option(
+    "--continue-on-error",
+    "-c",
+    is_flag=True,
+    default=False,
+    help="Continue the import if an asset fails to import. Should be used with --split",
+)
+@click.option(
     "--asset-type",
     type=click.Choice([rt.resource_name for rt in ResourceType], case_sensitive=False),
     callback=normalize_to_enum,
@@ -221,6 +230,7 @@ def native(  # pylint: disable=too-many-locals, too-many-arguments, too-many-bra
     external_url_prefix: str = "",
     load_env: bool = False,
     split: bool = False,
+    continue_on_error: bool = False,
     asset_type: Optional[ResourceType] = None,
 ) -> None:
     """
@@ -297,16 +307,17 @@ def native(  # pylint: disable=too-many-locals, too-many-arguments, too-many-bra
             configs["bundle" / relative_path] = config
 
     if split:
-        import_resources_individually(configs, client, overwrite)
+        import_resources_individually(configs, client, overwrite, continue_on_error)
     else:
         contents = {str(k): yaml.dump(v) for k, v in configs.items()}
         import_resources(contents, client, overwrite, asset_type=asset_type)
 
 
-def import_resources_individually(
+def import_resources_individually(  # pylint: disable=too-many-locals
     configs: Dict[Path, AssetConfig],
     client: SupersetClient,
     overwrite: bool,
+    continue_on_error: bool = False,
 ) -> None:
     """
     Import contents individually.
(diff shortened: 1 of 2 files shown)


src/preset_cli/cli/superset/sync/native/command.py
```diff
@@ -201,6 +203,13 @@ def render_yaml(path: Path, env: Dict[str, Any]) -> Dict[str, Any]:
     default=False,
     help="Split imports into individual assets",
 )
+@click.option(
+    "--continue-on-error",
+    "-c",
+    is_flag=True,
+    default=False,
+    help="Continue the import if an asset fails to import. Should be used with --split",
```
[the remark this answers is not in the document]
Vitor-Avila: oh, great catch! Let me do that change.

src/preset_cli/cli/superset/sync/native/command.py
```diff
@@ -297,59 +307,85 @@ def native(  # pylint: disable=too-many-locals, too-many-arguments, too-many-bra
...
+    imports = [
+        ("databases", lambda config: []),
+        ("datasets", lambda config: [config["database_uuid"]]),
+        ("charts", lambda config: [config["dataset_uuid"]]),
+        ("dashboards", get_dashboard_related_uuids),
+    ]
+    asset_configs: Dict[Path, AssetConfig]
+    related_configs: Dict[str, Dict[Path, AssetConfig]] = {}
+
+    file_path = Path("checkpoint.log" if not continue_on_error else "failures.log")
```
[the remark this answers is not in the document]
Vitor-Avila: Alright -- Implemented this in a new commit.

src/preset_cli/cli/superset/sync/native/command.py
```diff
@@ -232,6 +242,9 @@ def native(  # pylint: disable=too-many-locals, too-many-arguments, too-many-bra
     root = Path(directory)
     base_url = URL(external_url_prefix) if external_url_prefix else None
 
+    # The ``--continune-on-error`` flag should force the split option
```
betodealmeida, line 245: Typo: `s/continune/continue/`

src/preset_cli/cli/superset/sync/native/command.py
```diff
@@ -297,59 +310,115 @@ def native(  # pylint: disable=too-many-locals, too-many-arguments, too-many-bra
...
         ]
-        related_configs: Dict[str, Dict[Path, AssetConfig]] = {}
-        for resource_name, get_related_uuids in imports:
-            for path, config in configs.items():
-                if path.parts[1] != resource_name:
-                    continue
+    else:
+        logs["assets"] = []
+
+    assets_to_skip = {Path(log["path"]) for log in logs["assets"]} if logs else set()
```
betodealmeida, line 360: The `if logs else set()` is not needed, since `logs` will always have the `assets` key (set in line 354 or 358), and hence will always be truthy.

src/preset_cli/cli/superset/sync/native/command.py
```diff
@@ -297,59 +310,115 @@ def native(  # pylint: disable=too-many-locals, too-many-arguments, too-many-bra
...
+    """
+    Adds a log entry to the logs to skip future occurrences.
+    """
+    log_entry = {
+        "path": str(asset_path),
+        "status": status,
+        "uuid": asset_uuid,
+    }
+    if "assets" not in logs:
+        logs["assets"] = []
```
betodealmeida, line 411: It's kind of old-school, but you can do this in 1 line:

```python
logs.setdefault("assets", [])
```

betodealmeida (approved):
Double-approve!

src/preset_cli/cli/superset/sync/native/command.py
Vitor-Avila: ooops 🤦 fixing it now!

src/preset_cli/cli/superset/sync/native/command.py
Vitor-Avila: ohh good catch! I added that enforcement after having this, and forgot to remove it -- fixing it now too.

src/preset_cli/cli/superset/sync/native/command.py
Vitor-Avila: hmm I'm actually thinking I might not need this as well. Going to remove it.

PR #330, opened 2024-12-16
fix(dbt Cloud): Allow None on custom_branch_only field and update some endpoints to v3

Vitor-Avila:
When loading information about jobs from a dbt Cloud project, it's possible that the `custom_branch_only` field (from the `TriggerSchema`) returns `None`. This PR adds supports for such scenario.

It also updates some endpoints to use [dbt Cloud v3 API](https://docs.getdbt.com/docs/dbt-cloud-apis/admin-cloud-api).

### DIFF
--- a/src/preset_cli/api/clients/dbt.py
+++ b/src/preset_cli/api/clients/dbt.py
@@ -423,7 +423,7 @@ class TriggerSchema(PostelSchema):
     github_webhook = fields.Boolean(required=True)
     git_provider_webhook = fields.Boolean()
     schedule = fields.Boolean(required=True)
-    custom_branch_only = fields.Boolean()
+    custom_branch_only = fields.Boolean(allow_none=True)
 
 
 class SettingsSchema(PostelSchema):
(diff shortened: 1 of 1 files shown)


src/preset_cli/api/clients/dbt.py
```diff
@@ -724,7 +724,7 @@ def get_accounts(self) -> List[AccountSchema]:
         """
         List all accounts.
         """
-        url = self.baseurl / "api/v2/accounts/"
+        url = self.baseurl / "api/v3/accounts/"
```
betodealmeida, line 727: Maybe we could define somewhere:

```python
V2_API = self.baseurl / "api/v2"
V3_API = self.baseurl / "api/v3"
```

Then here we could:

```python
url = V3_API / "accounts/"
```

betodealmeida (approved):
giphy-4201342743

src/preset_cli/api/clients/dbt.py
Vitor-Avila: ohh, that's a good idea! I have to do a follow up on other dbt Cloud related issue and I can include this.

PR #331, opened 2025-01-03
feat(import ownership): Continue on error + skip import failures

Vitor-Avila:
https://github.com/preset-io/backend-sdk/pull/328 added the ability to complete an import if errors happen. This PR extends this logic to the ``import-ownership`` command:
* If a failure happens, it gets logged and the import continues.
* It also skips any asset that was failed to import with the ``import-assets`` command.

Some of the logic is redundant to both commands, so it got moved to a new ``lib.py`` file.

### DIFF
--- a/src/preset_cli/cli/superset/import_.py
+++ b/src/preset_cli/cli/superset/import_.py
@@ -7,6 +7,12 @@ import yaml
 from yarl import URL
 
 from preset_cli.api.clients.superset import SupersetClient
+from preset_cli.cli.superset.lib import (
+    add_asset_to_log_dict,
+    clean_logs,
+    get_logs,
+    write_logs_to_file,
+)
 
 
 @click.command()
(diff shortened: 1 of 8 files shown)


src/preset_cli/cli/superset/lib.py
```diff
@@ -0,0 +1,77 @@
...
+
+from pathlib import Path
+from typing import Any, Dict, Set
+
+import yaml
+
+LOG_FILE_PATH = Path("progress.log")
+
+
+def get_logs() -> dict[str, Any]:
```
betodealmeida, line 15: Yeah, let's create a `LogType` (or something) and return it here, with owners and assets set to empty lists if not present.

betodealmeida (approved):
Looks great! I left a few suggestions, but we can address them later it you need to merge this ASAP.

Vitor-Avila:
@betodealmeida these are all super valid suggestions. I'll merge this as-is now, but I'll get back to this in another PR.

src/preset_cli/cli/superset/import_.py
```diff
@@ -57,16 +63,73 @@ def import_roles(ctx: click.core.Context, path: str) -> None:
...
     auth = ctx.obj["AUTH"]
     url = URL(ctx.obj["INSTANCE"])
     client = SupersetClient(url, auth)
 
+    logs = get_logs()
+    failed_assets = (
+        {log["uuid"] for log in logs["assets"] if log["status"] == "FAILED"}
+        if logs.get("assets")
+        else set()
+    )
```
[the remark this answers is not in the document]
Vitor-Avila: Implemented here: https://github.com/preset-io/backend-sdk/pull/335

src/preset_cli/cli/superset/lib.py
Vitor-Avila: Covered here https://github.com/preset-io/backend-sdk/pull/335

PR #332, opened 2025-01-06
feat(dbt sync): Sync calculated columns from dbt

Vitor-Avila:
This PR adds support for syncing calculated columns from the dbt model. These should be defined under `$model.meta.superset.calculated_columns` (list).

The behavior is:

**Default sync (`reload_columns` set to `True` and `merge_metadata` set to `False`)**

In this configuration, metrics from dbt are synced to Superset, and metrics that were only present in the dataset (and not in the model) are deleted. This is the same behavior for calculated columns:
* If there isn't any calculated columns defined in dbt, then existing calculated columns from the dataset are preserved. This is to avoid breaking changes (as currently calculated columns aren't deleted in this flow).
* If there's any calculated column defined in dbt, then these are synced and Superset-only calculated columns are removed. 

**Merge metadata (`merge_metadata` set to `True` and `reload_columns` set to `False`)**
* Superset-only calculated columns are preserved.
* Metadata defined in dbt is synced.

**Preserve Metadata (both `merge_metadata` and `reload_columns` set to `False`)**
* Superset is the source of truth.
* new dbt calculated columns are synced.

### DIFF
--- a/src/preset_cli/cli/superset/sync/dbt/datasets.py
+++ b/src/preset_cli/cli/superset/sync/dbt/datasets.py
@@ -268,19 +268,19 @@ def compute_columns_metadata(  # pylint: disable=too-many-branches, too-many-arg
             dict_merge(final_column, dbt_metadata[column])
             dbt_metadata[column] = final_column
 
-    dbt_calc_columns_ = {c["column_name"]: c for c in dbt_calc_columns}
+    dbt_calc_columns_by_name = {c["column_name"]: c for c in dbt_calc_columns}
 
     if column_defaults:
-        for column, definition in dbt_calc_columns_.items():
+        for column, definition in dbt_calc_columns_by_name.items():
             final_column = copy.deepcopy(column_defaults)
             dict_merge(final_column, definition)
-            dbt_calc_columns_[column] = final_column
-    if reload_columns and dbt_calc_columns_:
+            dbt_calc_columns_by_name[column] = final_column
+    if reload_columns and dbt_calc_columns_by_name:
         dataset_columns = [
             column
             for column in dataset_columns
             if not column.get("expression")
-            or column["column_name"] in dbt_calc_columns_
+            or column["column_name"] in dbt_calc_columns_by_name
         ]
 
     for column in dataset_columns:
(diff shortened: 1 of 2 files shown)


src/preset_cli/cli/superset/sync/dbt/datasets.py
```diff
@@ -267,12 +268,34 @@ def compute_columns_metadata(
             dict_merge(final_column, dbt_metadata[column])
             dbt_metadata[column] = final_column
 
+    dbt_calc_columns_ = {c["column_name"]: c for c in dbt_calc_columns}
```
Vitor-Avila, line 271: Tried using the same var name (`dbt_calc_columns = {c["column_name"]: c for c in dbt_calc_columns}`) but `mypy` was going crazy so had to use this.

betodealmeida: Yeah, from a pure point of view the type of a variable should never change, and in that case it would change from `list` to `dict`.

Maybe `dbt_calc_columns_by_name`, if it's not too big?

src/preset_cli/cli/superset/sync/dbt/datasets.py
```diff
@@ -403,14 +436,15 @@ def sync_datasets(  # pylint: disable=too-many-locals, too-many-arguments
             continue
 
         # update column metadata
-        if dbt_columns := model.get("columns"):
+        if (dbt_columns := model.get("columns")) or calculated_columns:
```
betodealmeida, line 439: Super nit, but I find this hard to read. I would split it into 2 lines:

```suggestion
        dbt_columns = model.get("columns")
        if dbt_columns or calculated_columns:
```

betodealmeida (approved):

src/preset_cli/cli/superset/sync/dbt/datasets.py
Vitor-Avila: great suggestion!

src/preset_cli/cli/superset/sync/dbt/datasets.py
Vitor-Avila: sure! gonna change that too

adamcunnington-mlg:
@Vitor-Avila please can you update description of this MR to show an example valid YAML specification of a computed column. I can see from the code that the CLI is just passing it through - but it's going through several functions so difficult for me to see the exact right spec. I can also see that the docs have not been updated... https://docs.preset.io/docs/dbt-sync

I'm assuming an example is like this:
```yaml
models:
- name: foo
  meta:
    superset:
      calculated_columns:
      - name: test
        filterable: true
        groupby: true
        expression: CONCAT("foo", "bar")
```
?

Vitor-Avila:
Hey @adamcunnington-mlg,

You're correct that all keys are just passed through during the sync, so this is not really a CLI configuration, but it actually has to be in accordance to the Superset API schema for a column:
image

Note that most of these are non-required, and when using `--merge-metadata` we'll use the `column_name` to decide if it's a new/existing column. For example, `id` and `uuid` would be generated if you're creating a new column. In the next sync, the column already exists so we'll find it by its `column_name`, then re-use the `id` and `uuid` in the update event.

adamcunnington-mlg:
@Vitor-Avila thanks - can you confirm the following is a valid config - e.g. all property names and structure are valid?

```yml
models:
- name: foo
  meta:
    superset:
      calculated_columns:
      - column_name: test
        filterable: true
        groupby: true
        expression: CONCAT("foo", "bar")
        verbose_name: Test
```

Vitor-Avila:
@adamcunnington-mlg yup that looks correct. Let me know if you face any errors

PR #333, opened 2025-01-06
chore: Remove the --preserve-columns flag

Vitor-Avila:
This PR removes the `--preserve-columns` flag, as it's been re-named to `--preserve-metadata` in https://github.com/preset-io/backend-sdk/pull/238.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -74,7 +74,7 @@ repos:
 #  - id: reorder-python-imports
 #    args: [--application-directories=.:src]
 - repo: https://github.com/hadialqattan/pycln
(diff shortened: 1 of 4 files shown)


betodealmeida (approved):
I think we might need to freeze `pylint` in `setup.cfg` to `2.15.5`?

Vitor-Avila:
> I think we might need to freeze `pylint` in `setup.cfg` to `2.15.5`?

Great catch. Let me do that.

Vitor-Avila:
That didn't work. I think we need to bump `pycln` to https://github.com/hadialqattan/pycln/releases/tag/v2.5.0. Just did that -- let's see 🤞

PR #337, opened 2025-01-21
feat: Support the --asset-type argument for individual imports

Vitor-Avila:
This PR adds support for the `--asset-type` selector for individual imports. This allows users to trigger an import for a specific asset type (including dependencies) that get imported via the asset type endpoint.

It also fixes a logic gap that was introduced in https://github.com/preset-io/backend-sdk/pull/328 -- while it's correct that successful files added to the `progress.log` should be skipped, they should still be added to the `assets_config` dict as they might be dependencies to existing assets.

### DIFF
--- a/src/preset_cli/cli/superset/sync/native/command.py
+++ b/src/preset_cli/cli/superset/sync/native/command.py
@@ -14,7 +14,7 @@ from enum import Enum
 from io import BytesIO
 from pathlib import Path
 from types import ModuleType
-from typing import Any, Dict, Iterator, Optional, Set, Tuple
+from typing import Any, Dict, Iterator, Set, Tuple
 from zipfile import ZipFile
 
 import backoff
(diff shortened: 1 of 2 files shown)


src/preset_cli/cli/superset/sync/native/command.py
```diff
@@ -354,10 +362,20 @@ def import_resources_individually(  # pylint: disable=too-many-locals
...
         for resource_name, get_related_uuids in imports:
             for path, config in configs.items():
-                if path.parts[1] != resource_name or path in assets_to_skip:
+                if path.parts[1] != resource_name:
                     continue
 
                 asset_configs = {path: config}
+                for uuid in get_related_uuids(config):
+                    asset_configs.update(related_configs[uuid])
+                related_configs[config["uuid"]] = asset_configs
```
Vitor-Avila, line 371: This was already there -- we just need to make sure it's executed before the asset is skipped to be imported.

betodealmeida (approved):

PR #339, opened 2025-01-28
feat(dbt Cloud): Specify the DB ID in the command

Vitor-Avila:
Currently the dbt Cloud sync command expects that the DB connection in Preset is created based on the DB name in dataset (the actual database name). Keeping these matched is non intuitive, specially now that Superset supports multi-catalog (meaning that a single DB connection in Superset can interact with n catalogs).

This PR introduces the ability to specify a `--database-id=x` in the dbt Cloud command to perform the mapping.

### DIFF
--- a/src/preset_cli/cli/superset/sync/dbt/command.py
+++ b/src/preset_cli/cli/superset/sync/dbt/command.py
@@ -2,11 +2,13 @@
 A command to sync dbt models/metrics to Superset and charts/dashboards back as exposures.
 """
 
+from __future__ import annotations
+
 import logging
 import os.path
 import subprocess
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Tuple
+from typing import Any, Dict, List, Tuple
 
 import click
 import yaml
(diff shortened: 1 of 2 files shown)


betodealmeida (approved):
Looks good, but wouldn't it be better to use the UUID instead? (I'm not sure if it's easy.)

Vitor-Avila:
> Looks good, but wouldn't it be better to use the UUID instead? (I'm not sure if it's easy.)

I think the `uuid` would be easier if the user is doing both the `native` and `dbt` sync (as they would have the YAML handy and that's a universal ID even across Workspaces). But for the dbt sync specifically, I think the `id` or the `database_name` would be easier (considering they don't have the YAML and the `uuid` is not exposed on a lot of endpoints). 

If you think it makes more sense I could update this to be `--database-name` instead (or even add both and make them exclusive (you can use one OR the other at a time)).

betodealmeida:
> I think the uuid would be easier if the user is doing both the native and dbt sync (as they would have the YAML handy and that's a universal ID even across Workspaces). But for the dbt sync specifically, I think the id or the database_name would be easier (considering they don't have the YAML and the uuid is not exposed on a lot of endpoints).

Ah, makes sense!

betodealmeida:
Nice! Double approved! ✅✅

PR #340, opened 2025-01-29
feat(import assets): Pass DB passwords in the import command

Vitor-Avila:
Currently when importing a new DB connection the CLI prompts the user to enter its password. This PR adds the ability to specify the DB password as an argument to the CLI command, to skip the prompt (better for automations/integrations).

### DIFF
--- a/src/preset_cli/cli/superset/sync/native/command.py
+++ b/src/preset_cli/cli/superset/sync/native/command.py
@@ -278,11 +278,7 @@ def native(  # pylint: disable=too-many-locals, too-many-arguments, too-many-bra
     if load_env:
         env["env"] = os.environ  # type: ignore
 
-    pwds = (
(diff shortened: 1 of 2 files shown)


src/preset_cli/cli/superset/sync/native/command.py
```diff
@@ -272,6 +278,12 @@ def native(  # pylint: disable=too-many-locals, too-many-arguments, too-many-bra
     if load_env:
         env["env"] = os.environ  # type: ignore
 
+    pwds = (
+        {kv.partition("=")[0]: kv.partition("=")[2] for kv in db_password}
+        if db_password
+        else {}
+    )
```
betodealmeida, line 285:
```suggestion
    pwds = dict(kv.split("=", 1) for kv in db_password or [])
```

betodealmeida (approved):
Nice!

PR #342, opened 2025-02-03
chore: Add debug logging for import resources individually

Vitor-Avila:
Currently we're logging when importing a file, but we still iterate through other files (dependencies) that don't get imported, and an issue could occur during this step. This additional `debug` statement helps troubleshooting this issue.

### DIFF
--- a/tests/cli/superset/sync/native/command_test.py
+++ b/tests/cli/superset/sync/native/command_test.py
@@ -6,7 +6,7 @@ Tests for the native import command.
 
 import json
 from pathlib import Path
-from typing import Dict, List
+from typing import Any, Dict, List
 from unittest import mock
 from zipfile import ZipFile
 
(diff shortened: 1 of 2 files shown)


tests/cli/superset/sync/native/command_test.py
```diff
@@ -1536,6 +1536,87 @@ def test_import_resources_individually_continue(
     assert not Path("progress.log").exists()
 
 
+def test_import_resources_individually_debug(mocker: MockerFixture) -> None:
```
Antonio-RiveroMartnez, line 1539: Just an observation, we should not be testing against logger calls since that's an implementation detail and are prone to break if we ever change the content of the message, remove it etc.

Antonio-RiveroMartnez (approved):

tests/cli/superset/sync/native/command_test.py
Vitor-Avila: I think your comment is totally fair. I added the test here as well just to ensure future changes won't break this logic -- since this is a CLI, users rely on following the progress in the terminal, so we should make sure details are logged. This was added as a `debug` statement, so it's only going to be visible to customers if they trigger the CLI command with `--loglevel=DEBUG`.

PR #347, opened 2025-02-28
fix(Preset): Export users

Vitor-Avila:
In Preset, the Superset Users List view is disabled (since users are managed through Preset Manager). That said, in order to export users (with their local Superset IDs) we would parse the DOM of the Add DAR view. 

There was a change to this view so that the users are not loaded to the DOM anymore -- instead, there's a user search field. 

This PR updates the logic specifically for Preset Cloud to get the user information from the `/api/v1/chart/related/owners` field (which lists all members with proper pagination).

This is a temporary fix, as there's ongoing work to migrate the Superset Roles view to RESTful, and later do the same with Preset's DAR view.

### DIFF
--- a/src/preset_cli/api/clients/preset.py
+++ b/src/preset_cli/api/clients/preset.py
@@ -7,7 +7,7 @@ import logging
 from enum import Enum
 from typing import Any, Dict, Iterator, List, Optional, Union
 
-from bs4 import BeautifulSoup
+import prison
 from yarl import URL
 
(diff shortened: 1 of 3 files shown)


Vitor-Avila:
Also fixes https://github.com/preset-io/backend-sdk/pull/347

Antonio-RiveroMartnez (approved):
LGTM! This is a tmp fix since we're working on the Role CRUD RESTful migration

PR #349, opened 2025-03-19
fix: allow empty project ID in connection

betodealmeida:
Looks like `project_id` can be null:

```python
{
  "status": {
    "code": 200,
    "is_success": true,
    "user_message": "Success!",
    "developer_message": ""
  },
  "data": [
    {
      "id": 70471823450461,
      "name": "Analytics",
      ...
      "connection": {
        "id": 70471823416830,
        "account_id": 70471823445028,
        "project_id": null,  # <== HERE
        ...
      },
      "environments": null,
      "repository": {...},
      "group_permissions": [],
      "docs_job": null,
      "freshness_job": null
    }
  ],
  "extra": {...},
  "error_code": null
}
```

This PR removes all unused fields from the Marshmallow schemas, so that if they are different from what we expect nothing breaks. It also moves the logic of confirming differences between dbt Core and Cloud into the schema — eg, a response having `depends_on` and another having `dependsOn`.

I tested it with this script as well:

```python
from pprint import pprint

from preset_cli.auth.token import TokenAuth
from preset_cli.api.clients.dbt import DBTClient


auth = TokenAuth("dbtc_XXX")
client = DBTClient(auth, access_url="https://bb691.us1.dbt.com/")

accounts = client.get_accounts() print(len(accounts))
account = accounts[0]

projects = client.get_projects(account["id"])
print(len(projects))
project = projects[0]

jobs = client.get_jobs(account["id"], project["id"])
print(len(jobs))
job = jobs[0]

models = client.get_models(job["id"])
print(len(models))
pprint(models)
```

This was initially failing, and now it works.


src/preset_cli/api/clients/dbt.py
```diff
@@ -514,53 +87,54 @@ class JobSchema(PostelSchema):
...
     name = fields.String()
-    unique_id = fields.String(data_key="uniqueId")
+    alias = fields.String(allow_none=True)
+    unique_id = fields.String()
     tags = fields.List(fields.String())
     columns = fields.Raw(allow_none=True)
     config = fields.Dict(fields.String(), fields.Raw(allow_none=True))
 
+    @pre_load
+    def rename_fields(  # pylint: disable=unused-argument
```
Vitor-Avila, line 112: Super cool!

src/preset_cli/api/clients/dbt.py
```diff
@@ -514,53 +87,54 @@ class JobSchema(PostelSchema):
...
+        if isinstance(columns := data.get("columns"), dict):
+            data["columns"] = list(columns.values())
+
+        if "dependsOn" in data:
+            data["depends_on"] = data.pop("dependsOn")
+        depends_on = data.get("depends_on", [])
+
+        if isinstance(depends_on, dict):
+            data["depends_on"] = depends_on["nodes"]
+
```
Vitor-Avila, line 135: Do we need this for `ModelSchema` too? I see down there that this logic was implemented in `get_og_metric_from_config` (before loading the metric to `OGMetricSchema`) but curious if we also need to do this for model.

src/preset_cli/cli/superset/sync/dbt/command.py
```diff
@@ -189,10 +189,8 @@ def dbt_core(  # pylint: disable=too-many-arguments, too-many-branches, too-many
     models = []
     for config in configs["nodes"].values():
         if config["resource_type"] == "model":
-            # conform to the same schema that dbt Cloud uses for models
-            unique_id = config["uniqueId"] = config["unique_id"]
+            unique_id = config["unique_id"]
             config["children"] = configs["child_map"][unique_id]
```
Vitor-Avila, line 193: Should this `config["children"] = configs["child_map"][unique_id]` be handled as part of the `@pre_load` logic?

betodealmeida: I considered that, but note that `child_map` lives in the manifest root, not inside the model config, so `pre_load` doesn't have access to it.

src/preset_cli/api/clients/dbt.py
betodealmeida: I think so, this is from the manifest:

```python
  "nodes": {
    "model.superset_examples.messages_channels": {
      "raw_sql": "...",
      "compiled": true,
      "resource_type": "model",
      "depends_on": {
        "macros": [],
        "nodes": [
          "source.superset_examples.public.channels",
          "source.superset_examples.public.messages"
        ]
      },
```

src/preset_cli/cli/superset/sync/dbt/command.py
Vitor-Avila: ohhh, good call! Did not notice that.

src/preset_cli/api/clients/dbt.py
Vitor-Avila: Gotcha! Thanks for confirming it

PR #356, opened 2025-07-02
chore: Support getting meta.superset from config

Vitor-Avila:
dbt is moving the `meta` properly (previously a top-level property) to inside the `config` property:
* https://docs.getdbt.com/reference/deprecations#propertymovedtoconfigdeprecation
* https://github.com/dbt-labs/dbt-core/issues/11651
* https://docs.getdbt.com/reference/resource-configs/meta

This PR adds support for this change (backwards compatibility should be preserved).


betodealmeida (approved):
Looks great!

PR #359, opened 2025-08-25
feat(dbt sync): Support using nested properties under `config` to filter models

Vitor-Avila:
This PR adds support for filtering dbt models using a nested config property (such as `config.meta.$key`).

### DIFF
--- a/src/preset_cli/cli/superset/sync/dbt/lib.py
+++ b/src/preset_cli/cli/superset/sync/dbt/lib.py
@@ -312,15 +312,14 @@ def load_profiles(
     return apply_templating(profiles)
 
 
-# pylint: disable=R0911
+# pylint: disable=R0911, R0912
 def filter_models(models: List[ModelSchema], condition: str) -> List[ModelSchema]:
     """
     Filter a list of dbt models given a select condition.
 
     Currently supports a subset of the dbt selector syntax:
     - tag:value - Filter by tag
-    - config.key:value - Filter by config property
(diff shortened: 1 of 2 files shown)


src/preset_cli/cli/superset/sync/dbt/lib.py
```diff
@@ -328,9 +331,18 @@ def filter_models(models: List[ModelSchema], condition: str) -> List[ModelSchema
...
     if condition.startswith("config"):
         filtered_models = []
-        config_key, config_value = re.split(r"[.:]", condition)[1:]
+        config_key, _, config_value = condition.rpartition(":")
+        parts = config_key.split(".")
         for model in models:
-            if model.get("config", {}).get(config_key) == config_value:
+            config = model.get("config", {})
+
+            if len(parts) == 2:  # config.key:value
```
betodealmeida, line 339: What about `config.foo.bar.baz`? We could make this work with any number of parts, something like:

```python
value = model
while parts:
    part = parts.pop(0)
    value = value.get(part, {})

if value == config_value:
    filtered_models.append(model)
```

betodealmeida (commented):
Are more than 2 levels of nesting allowed? If so we should future-proof it by supporting n levels.

src/preset_cli/cli/superset/sync/dbt/lib.py
Vitor-Avila: alright, done! 🙌

betodealmeida (approved):
Thanks!!

PR #364, opened 2025-10-03
feat: Local overwrite based on uuid

Vitor-Avila:
The export command validates if the exported assets already exist in the target directory, raising if any asset already exists and the `overwrite` flag is not included. The current validation is solely based on the file name, which is not ideal as assets can be renamed in Superset, leading to duplicated files in the directory, and causing a race condition during import (the one imported last reflects in the UI).

This PR updates the logic to first check by file name, and then checks by `uuid`.

### DIFF
--- a/src/preset_cli/cli/superset/export.py
+++ b/src/preset_cli/cli/superset/export.py
@@ -6,7 +6,7 @@ import json
 import re
 from collections import defaultdict
 from pathlib import Path
-from typing import Any, Callable, List, Set, Tuple, Union
+from typing import Any, Callable, Dict, List, Optional, Set, Tuple, Union
 from zipfile import ZipFile
 
 import click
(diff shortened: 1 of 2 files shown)


src/preset_cli/cli/superset/export.py
```diff
@@ -26,6 +26,108 @@ def get_newline_char(force_unix_eol: bool = False) -> Union[str, None]:
...
+    Check if the asset is new to the directory or not.
+    """
+    # First validate by file name
+    if file_path.exists() and not overwrite:
+        raise Exception(
+            f"File already exists and ``--overwrite`` was not specified: {file_path}",
+        )
+    # Short circuit as file will be automatically overwritten based on its filename
+    if file_path.exists():
+        return
```
betodealmeida, line 98:
```suggestion
    if file_path.exists():
        if not overwrite:
            raise Exception(
                f"File already exists and ``--overwrite`` was not specified: {file_path}",
            )
        return
```

betodealmeida (approved):
Nice! I like how the old file gets deleted when the asset is renamed!

PR #367, opened 2025-12-16
feat: OSI sync

betodealmeida:
Command to sync from the OSI spec (https://github.com/open-semantic-interchange/OSI/blob/main/core-spec/spec.md) to Superset:

```bash
$ preset-cli --workspaces=https://49437123.us1a.app.preset.io superset sync osi --database-id=46 tpcds_semantic_model.yaml
```

It will create physical datasets for the models, and add the corresponding metrics. It also creates a denormalized virtual dataset with all the metrics. For example:

```sql
SELECT
  "SS_QUANTITY" AS "SS_QUANTITY",
  SUM(ss_ext_sales_price) AS "total_sales"
FROM (
  SELECT
    store_sales.*,
    date_dim.*,
    customer.*,
    item.*,
    store.*
  FROM SAMPLE_DATA.tpcds_sf10tcl.store_sales AS store_sales
  LEFT JOIN SAMPLE_DATA.tpcds_sf10tcl.date_dim AS date_dim
    ON store_sales.ss_sold_date_sk = date_dim.d_date_sk
  LEFT JOIN SAMPLE_DATA.tpcds_sf10tcl.customer AS customer
    ON store_sales.ss_customer_sk = customer.c_customer_sk
  LEFT JOIN SAMPLE_DATA.tpcds_sf10tcl.item AS item
    ON store_sales.ss_item_sk = item.i_item_sk
  LEFT JOIN SAMPLE_DATA.tpcds_sf10tcl.store AS store
    ON store_sales.ss_store_sk = store.s_store_sk
) AS virtual_table
WHERE
  "D_DATE" >= CAST('2002-12-01T00:00:00.000000' AS DATETIME)
  AND "D_DATE" < CAST('2002-12-31T00:00:00.000000' AS DATETIME)
GROUP BY
  "SS_QUANTITY"
ORDER BY
  "total_sales" DESC
LIMIT 10000;
```

Snowflake's query optimizer should perform join elimination for the unused tables, so the query shouldn't be as bad as it looks. In this query:

- `date_dim` - NEEDED (used in `WHERE` clause for `D_DATE` filter)
- `customer` - NOT NEEDED (no columns used)
- `item` - NOT NEEDED (no columns used)
- `store` - NOT NEEDED (no columns used)

Snowflake can eliminate LEFT JOINs when:

1. No columns from the right table are in SELECT, WHERE, GROUP BY, or ORDER BY
2. The join doesn't affect row cardinality (LEFT JOINs preserve all left-side rows)


src/preset_cli/cli/superset/sync/dj/command.py
```diff
@@ -70,6 +69,10 @@ def dj(  # pylint: disable=invalid-name,too-many-arguments
     """
     Sync DJ cubes to Superset.
     """
+    # Lazy import to avoid ImportError when datajunction is not installed
+    # pylint: disable=import-outside-toplevel,no-name-in-module
```
Vitor-Avila: Nice!

src/preset_cli/cli/superset/sync/osi/command.py
```diff
@@ -0,0 +1,109 @@
...
+
+    # Validate options
+    if preserve_metadata and merge_metadata:
+        raise click.ClickException(
+            "--preserve-metadata and --merge-metadata cannot be combined. "
+            "Please use only one.",
+        )
+
+    # Parse OSI file
+    click.echo(f"Parsing OSI file: {file}")
```
Vitor-Avila: This PR includes a few `click.echo()` statements. Does it make sense to keep these all? I wonder if we should move the progress-related ones to `logger.debug()` so that users can still see them if desired, but the command is quieter by default.

src/preset_cli/cli/superset/sync/osi/lib.py
```diff
@@ -0,0 +1,970 @@
...
+    "athena": "presto",  # Athena uses Presto dialect
+    "awsathena+rest": "presto",
+}
+
+# OSI dialect names to sqlglot dialect names
+OSI_DIALECT_MAP = {
+    "ANSI_SQL": None,  # sqlglot default
+    "SNOWFLAKE": "snowflake",
+    "MDX": None,  # Not supported by sqlglot
+    "TABLEAU": None,  # Not supported by sqlglot
```
Vitor-Avila: 🔫

src/preset_cli/cli/superset/sync/osi/lib.py
```diff
@@ -0,0 +1,970 @@
...
+    Get or create a physical dataset for an OSI dataset definition.
+    """
+    source = parse_source(osi_dataset["source"])
+    table_name = source["table"]
+    schema = source["schema"]
+
+    # Check if dataset already exists
+    filters: Dict[str, Any] = {
+        "database": OneToMany(database["id"]),
+        "table_name": table_name,
```
Vitor-Avila: should we filter by `catalog` too? Can't remember if it's supported.

src/preset_cli/cli/superset/sync/osi/lib.py
```diff
@@ -0,0 +1,970 @@
...
+    superset_metric: SupersetMetricDefinition = {
+        "metric_name": name,
+        "expression": cleaned_expression,
+        "metric_type": "sql",
+        "verbose_name": name.replace("_", " ").title(),
+    }
+
+    description = metric.get("description")
+    if description:
+        superset_metric["description"] = description
```
Vitor-Avila: We can likely use walrus here

src/preset_cli/cli/superset/sync/osi/lib.py
```diff
@@ -0,0 +1,970 @@
...
+    if failures:
+        click.echo(
+            click.style(
+                f"\nWarning: {len(failures)} of {len(osi_datasets)} physical "
+                f"dataset(s) failed to sync:",
+                fg="yellow",
+            ),
+        )
+        for name, error in failures:
+            click.echo(click.style(f"  - {name}: {error}", fg="yellow"))
```
Vitor-Avila: We could probably incorporate the `raise_failures` logic here.

Vitor-Avila (commented):
Left non-blocking comments

PR #386, opened 2026-07-21
fix(API): Export roles

Vitor-Avila:
The Roles Export logic was still relying on crawling HTML, which doesn't work for newer Superset versions. This PR introduces the following changes:
* Renamed `export_users` to `get_users`.
* Removed Preset-specific logic to get users from a Workspace context (Preset now supports `/api/v1/security/users/`).
* Kept `_get_users_legacy` to crawl HTML for older Superset instances.
* `export_roles` now relies on either `_export_roles_superset` (via `/api/v1/security/permissions-resources`, `/api/v1/security/groups` and `/api/v1/security/roles/search`) or `_export_dars_preset` (via `/api/v1/security/dar`).
* `_export_roles_legacy` still preserved for older Superset instances.
* Added a helper for `_is_preset_workspace`.

### DIFF
--- a/src/preset_cli/api/clients/superset.py
+++ b/src/preset_cli/api/clients/superset.py
@@ -49,9 +49,9 @@ from bs4 import BeautifulSoup
 from yarl import URL
 
 from preset_cli import __version__
-from preset_cli.api.clients.preset import PresetClient
 from preset_cli.api.operators import Equal, In, Operator
 from preset_cli.auth.main import Auth
+from preset_cli.exceptions import SupersetError
 from preset_cli.lib import remove_root, validate_response
 from preset_cli.typing import UserType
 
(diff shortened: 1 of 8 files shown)


src/preset_cli/api/clients/superset.py
```diff
@@ -49,18 +49,17 @@
...
 from preset_cli.lib import remove_root, validate_response
 from preset_cli.typing import UserType
 
 _logger = logging.getLogger(__name__)
 
 MAX_PAGE_SIZE = 100
 MAX_IDS_IN_EXPORT = 50
-
-
+PRESET_WORKSPACE_DOMAIN = "preset.io"
```
Antonio-RiveroMartnez, line 62: nit: not only this but seeing a few configs value in the module, we could extract out?

src/preset_cli/api/clients/superset.py
```diff
@@ -750,11 +758,36 @@ def delete_dashboard(self, dashboard_id: int) -> Any:
         """
         self.delete_resource("dashboard", dashboard_id)
 
-    def get_users(self, **kwargs: str) -> List[Any]:
+    def get_users(self, **kwargs: str) -> List[UserType]:
```
Antonio-RiveroMartnez, line 761: ❤️

Antonio-RiveroMartnez (approved):

